PR #2, opened 2023-07-06
Initial BE skeleton setup

dimitrov93:
GET point to /home


dimitrov93:
Initial setup

.idea/workspace.xml
VladoKat, line 1: We should ignore some of .idea's contents. Please, refer to this article by JetBrains -> https://git-scm.com/book/en/v2/Getting-Started-The-Command-Line. They have mentioned a .gitignore file to use https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore

jewellery-inventory/HELP.md
VladoKat, line 1: Do we need this file? I imagine it's generated automatically.

jewellery-inventory/.vscode/launch.json
VladoKat, line 1: I think we don't need .vscode configuration in this project if we are using Intellij for the backend. Should we add it to the .gitignore file as well?

jewellery-inventory/target/classes/application.properties
VladoKat, line 1: We should ignore the target folder as well.  We don't want to include the built service in Github. -> https://github.com/github/gitignore/blob/main/Maven.gitignore

VladoKat (changes requested):
According to this PR, there are no project files added apart from the .idea, .vscode and target folder. As I've mentioned, we should actually ignore those folders (or parts of them) because otherwise they would pop up in every PR. Another thing is we would want to merge to the test branch first. By the way, let's have the package be `com.jewellery.inventory`.

PR #6, opened 2023-07-08
Feature 4 core entities

dimitrov93:


src/main/java/jewellery/inventory/model/Product.java
```diff
@@ -0,0 +1,108 @@
...
+    private List<ResourceInProduct> resourcesContent;
+
+    @ManyToMany
+    private List<Product> productsContent;
+
+    private String description;
+    private double salePrice;
+    private boolean isSold;
+
+    // ------ Getters and setters ------ //
```
VladoKat, line 35: Let's use [lombok ](https://projectlombok.org)to generate the boilerplate getters/setters/constructors

.gitignore
VladoKat, line 1: Something with this PR has gone wrong. .gitignore file, mvnw, etc. should already be there. There are currently conflicts as well. Please, pull main, branch from there and apply the model changes.

src/main/java/jewellery/inventory/model/User.java
```diff
@@ -0,0 +1,66 @@
...
+public class User {
+
+    @Id
+    @GeneratedValue
+    private UUID id;
+
+    private String name;
+    private String email;
+
+    @OneToMany(mappedBy = "user", cascade = CascadeType.ALL.ALL)
```
VladoKat, line 19: You have a bit of a typo in CascadeType.ALL.ALL. Second ALL is not needed. I am a bit surprised the IDE didn't complain about it. 
To be fair, I would remove the cascade type for User's fields for now because with CascadeType.ALL when we delete a user, it's going to delete all the products associated with the user and we don't want to do that.

src/main/java/jewellery/inventory/model/resources/Resource.java
```diff
@@ -0,0 +1,74 @@
...
+
+@Entity
+public class Resource {
+
+    @Id
+    @GeneratedValue
+    private UUID id;
+
+    private String name;
+    private String color;
```
VladoKat, line 15: As per the the [Data model](https://docs.google.com/document/d/1WDE9nPtZZ5vfveHt5Qyax_VhY2mdQGbfcojNYt8iVP0/edit) the Resource doesn't have a color.

src/main/java/jewellery/inventory/model/resources/Pear.java
```diff
@@ -0,0 +1,46 @@
+package jewellery.inventory.model.resources;
+
+import javax.persistence.Entity;
+
+@Entity
+public class Pear extends Resource {
```
VladoKat, line 6: Typo in the class name 🍐. Should be Pear**l**

src/main/java/jewellery/inventory/model/resources/Pear.java
```diff
@@ -0,0 +1,46 @@
...
+    private String type;
+    private double size;
+    private String quality;
+    private String shape;
```
VladoKat, line 10: It's missing a color property.

src/main/java/jewellery/inventory/model/resources/PreciousMetal.java
```diff
@@ -0,0 +1,37 @@
+package jewellery.inventory.model.resources;
+
+import javax.persistence.Entity;
+
+@Entity
+public class PreciousMetal extends Resource {
+    private String type;
+    private int purity;
+    private String plating;
```
VladoKat, line 9: Also missing color property.

VladoKat (changes requested):
The changes overall look good. The key things to be changed are:
1. Branch from main and apply the model changes so that there are no conflicts and files that are not added as part of this PR don't show up in the differences.
2. Make the PR to test in stead of main
3. Use [lombok](https://projectlombok.org) to get rid of the boilerplate getters/setters code.
4. Either keep the color field in Resource and remove it from the rest of the model classes that extend it or remove it from Resource and add it to the classes that extend it (all but the LinkingPart). 
Overall the PR looks good! The entity relations seem to be correct. Great job! 🏅

PR #7, opened 2023-07-09
Adding core entities + lombok + color has been removed from Resources

dimitrov93:


src/main/java/jewellery/inventory/model/resources/Pear.java
```diff
@@ -0,0 +1,14 @@
+package jewellery.inventory.model.resources;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class Pear extends Resource {
```
VladoKat: Please, rename this to Pearl

src/main/java/jewellery/inventory/JewelleryInventoryApplication.java
```diff
@@ -8,6 +8,6 @@ public class JewelleryInventoryApplication {
 
 	public static void main(String[] args) {
 		SpringApplication.run(JewelleryInventoryApplication.class, args);
-	}
 
+	}
```
VladoKat: Can you try to reset this file to the state it was? We don't want to have obsolete changes in the commits. You can try to do `git checkout origin/main src/main/java/jewellery/inventory/JewelleryInventoryApplication.java
` this will reset just the JewelleryInventoryApplication file to the state it was in main.

VladoKat (changes requested):
It looks good now. Can you just open the PR for the `test` branch?

VladoKat (approved):
Nice Job!

PR #8, opened 2023-07-10
Setup docker container with postgres db

Alekseeybg:
https://trello.com/c/zirxRk0X/5-be-setup-db


docker-entrypoint-initdb.d/init.sql
VladoKat, line 1: Do we need this file if it's empty?

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3.8'
+services:
+  db:
+    container_name: postgres
+    image: jms-db
```
VladoKat, line 5: The steps in https://docs.google.com/document/d/12QSq2K_E1DIsF0a99rClwVmwdu7eLKnVagnG0i9JjBI don't work because we don't have jms-db image created and this doesn't match any publicly created image from docker hub. Should this perhaps be the postgresql image from docker hub? 
Screenshot 2023-07-11 at 9 57 35

pom.xml
```diff
@@ -1,41 +1,60 @@
...
+    <name>jewellery-inventory</name>
+    <description>Project to help jewellery businesses digitalize their inventory management</description>
+    <properties>
+        <java.version>17</java.version>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
```
VladoKat: Let's make sure we are not changing formatting of previous code. We might want to introduce [google-java-formatting](https://stackoverflow.com/a/42979701/13062206) for intellij so that when each one of us formats the code, it doesn't end up changing something we haven't touched but let's do that as a separate PR.

src/main/java/jewellery/inventory/JewelleryInventoryApplication.java
```diff
@@ -6,8 +6,7 @@
 @SpringBootApplication
 public class JewelleryInventoryApplication {
 
-	public static void main(String[] args) {
-		SpringApplication.run(JewelleryInventoryApplication.class, args);
-	}
-
+    public static void main(String[] args) {
+        SpringApplication.run(JewelleryInventoryApplication.class, args);
+    }
```
VladoKat: Same here, you haven't made any changes here but it pops up as a change. Make sure we don't have obsolete changes.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+    restart: always
+    environment:
+      POSTGRES_DB: ${JMS_DATABASE_NAME}
+      POSTGRES_USER: ${JMS_DATABASE_USER}
+      POSTGRES_PASSWORD: ${JMS_DATABASE_PASSWORD}
```
VladoKat, line 13: What does JMS stand for? I saw it's Java Messaging Service but is that relevant to our app and the connection to the DB?

VladoKat (changes requested):
I couldn't start the DB due to an error that the image doesn't exist or it cannot be accessed. There is also a conflict in the pom.xml file. Let's make sure we take the latest changes from main (after the merge of https://github.com/JewelleryManagement/jewellery-management-service/pull/9)

src/main/java/jewellery/inventory/JewelleryInventoryApplication.java
Alekseeybg: I was messing with this file while testing but then reverted the changes I guess the formatting is the issue.

pom.xml
Alekseeybg: I added new dependencies here and I have a habit of pressing Alt + Ctrl + L

docker-entrypoint-initdb.d/init.sql
Alekseeybg: I removed it. I read somewhere it is a good practice to do that. But in our case we will probably not use it.

docker-compose.yml
Alekseeybg: Added Dockerfile so now the instructions should work.

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM postgres:13
+ARG POSTGRES_DB
+ARG POSTGRES_USER
+ARG POSTGRES_PASSWORD
+
+ENV POSTGRES_DB=${POSTGRES_DB}
+ENV POSTGRES_USER=${POSTGRES_USER}
+ENV POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
+
+ADD ./docker-entrypoint-initdb.d/init.sql /docker-entrypoint-initdb.d/
```
VladoKat: Now that we removed the file, this command fails while trying to build the Dockerfile. We can remove it for now as we don't have any initialization script.
Screenshot 2023-07-11 at 12 32 37

src/main/java/jewellery/inventory/JewelleryInventoryApplication.java
VladoKat: This is still showing up as change. Try to checkout the test version of it by doing `git checkout origin/test src/main/java/jewellery/inventory/JewelleryInventoryApplication.java`

docker-compose.yml
VladoKat: Do you know why we use JMS in the name?

VladoKat (changes requested):
Can you please, add a README file with a reference to the documentation you made?

docker-compose.yml
Alekseeybg: JMS = Jewellery Management System

Dockerfile
Alekseeybg: Removed the line. Also added a Readme file

src/main/java/jewellery/inventory/JewelleryInventoryApplication.java
Alekseeybg: Fixed

Alekseeybg:
> Can you please, add a README file with a reference to the documentation you made?

Added README file

VladoKat (approved):
Great Job! 🎖️

PR #13, opened 2023-07-18
Feature 8 expose user endpoint

Alekseeybg:


VladoKat:
https://trello.com/c/skyE8f3c/8-be-expose-user-endpoint

pom.xml
```diff
@@ -54,6 +67,28 @@
             <version>42.2.27</version>
         </dependency>
 
+        <dependency>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+        </dependency>
```
VladoKat: jakarta.validation-api comes implicitly with spring-boot-starter-validation. We don't need to add it explicitly. You can install [Maven Helper](https://plugins.jetbrains.com/plugin/7179-maven-helper) plugin in Intellij to help with dependency issues.

src/main/java/jewellery/inventory/constant/ApplicationConstants.java
VladoKat, line 1: I think we can get rid of this file altogether. Let's stick to the following rules:
1. If a constant is used only once, then don't create a constant. Keep it hardcoded.
2. If a constant is used in only one file, then move the constant in that file. 

From what I checked, if we follow these rules, there will be no need for the file as all the constants will either be moved to the file they are used in or they will be removed.

src/main/java/jewellery/inventory/controller/UserController.java
```diff
@@ -0,0 +1,56 @@
...
+
+@RestController
+@RequestMapping("/users")
+public class UserController {
+  @Autowired private UserService userService;
+
+  @GetMapping
+  public List<User> getAllUsers() {
+    return userService.getAllUsers();
+  }
```
VladoKat: Let's keep the responses consistent. Here we are returning a List<User> and in the next one we are using a ResponseEntity<User>.
We should also be using DTOs, not the entities themselves for the response.
We should also use different DTOs for request and response bodies.

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,102 @@
...
+          error.getDefaultMessage() != null
+              ? error.getDefaultMessage()
+              : "Error message not available";
+
+      if (!fieldErrorsMap.containsKey(fieldName)) {
+        fieldErrorsMap.put(fieldName, new ArrayList<>());
+      }
+
+      fieldErrorsMap.get(fieldName).add(errorMessage);
+    }
```
VladoKat, line 71: This takes some time to understand what it does exactly. You can extract it in a method and name it accordingly (i.e. getFieldErrors)

pom.xml
```diff
@@ -62,6 +97,32 @@
...
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.11.0</version>
+                <configuration>
+                    <source>${java.version}</source>
+                    <target>${java.version}</target>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.projectlombok</groupId>
+                            <artifactId>lombok</artifactId>
+                            <version>1.18.22</version>
```
VladoKat: You can also extract lombok's version as we are using it in 2 places.

src/test/java/jewellery/inventory/controller/UserControllerTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+    UUID updatingUserId = UUID.randomUUID();
+    User updatingUser = new User();
+    updatingUser.setId(updatingUserId);
+    updatingUser.setName("newUser");
+    updatingUser.setEmail("updating@example.com");
+    userService.createUser(updatingUser);
+
+    updatingUser.setName("John"); // duplicate name
+    updatingUser.setEmail("updating@example.com");
```
VladoKat: This part doesn't do anything because userService is mocked and we are not passing the updatingUser to the `when` method, right?

src/test/java/jewellery/inventory/service/UserServiceTest.java
```diff
@@ -0,0 +1,200 @@
...
+    verify(userRepository, times(1)).save(user);
+  }
+
+  @Test
+  void shouldThrowWhenUpdatingWithDuplicateName() {
+    UUID existingUserId = UUID.randomUUID();
+    User existingUser = new User();
+    existingUser.setId(existingUserId);
+    existingUser.setName("existing");
+    existingUser.setEmail("existing@example.com");
```
VladoKat: I see we are using the same hardcoded constants in several places. Let's extract them into constants.
Another suggestion would be to create the different users/dtos in a separate file and just call the methods here.

VladoKat (changes requested):
Implementation looks good overall. We just need to introduce DTOs in the controllers as response and request objects and not use the entities. Check the comments for other minor recommendations. I am open to discuss any of them.

pom.xml
Alekseeybg: I installed the plugin but it does not show any issues with jakarta-validation-api. It shows one issue with the snakeyaml but I put it on purpose because IntelliJ tells me about vulnerability in the version spring-boot-starter-web uses... Removed jakarta.validation-api dependency and the project seems to compile without issues.

pom.xml
Alekseeybg: Thank you for pointing that out. I made the proposed change.

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
Alekseeybg: Fixed.

src/main/java/jewellery/inventory/constant/ApplicationConstants.java
Alekseeybg: I see your point. There are constants used both in the service and in the tests though. For example USER_NOT_FOUND_ERROR_MSG and DUPLICATE_NAME_ERROR_MSG. I will remove the file and make the required changes

src/main/java/jewellery/inventory/controller/UserController.java
Alekseeybg: That is a very valid point. I made the requested changes.

pom.xml
VladoKat: yeah, I missaid issues. I meant it helps to see where different dependencies come form. In this case we had the jakarta-validation-api dependency twice and I saw it thanks to the plugin

src/main/java/jewellery/inventory/constant/ApplicationConstants.java
VladoKat: Yes, you are right they are used in 2 places. This made me think, though. Why don't we move the Error message constants to the Exceptions classes directly. We have 1 message per exception. It doesn't make sense to have them outside in this case.

src/test/java/jewellery/inventory/controller/UserControllerTest.java
Alekseeybg: Fixed that and added more tests

src/main/java/jewellery/inventory/constant/ApplicationConstants.java
Alekseeybg: I fixed the issue and deleted the constants file.

src/test/java/jewellery/inventory/service/UserServiceTest.java
Alekseeybg: I created separate file with methods creating mock data.

pom.xml
```diff
@@ -54,6 +69,30 @@
...
+            <groupId>org.mapstruct</groupId>
+            <artifactId>mapstruct-processor</artifactId>
+            <version>${org.mapstruct.version}</version>
+            <optional>true</optional>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok-mapstruct-binding</artifactId>
+            <version>0.2.0</version>
```
VladoKat: lombok-mapstruct-binding version is used also in the plugin path. Should we extract it as a property as well?

src/main/java/jewellery/inventory/controller/UserController.java
```diff
@@ -0,0 +1,66 @@
...
+@RestController
+@RequestMapping("/users")
+@CrossOrigin(origins = "http://localhost:3000")
+public class UserController {
+  @Autowired private UserService userService;
+
+  @GetMapping
+  public ResponseEntity<List<UserResponse>> getAllUsers() {
+    List<UserResponse> userResponseList =
+        UserMapper.INSTANCE.toUserResponseList(userService.getAllUsers());
```
VladoKat: Let's extract the mapping part in the service. Make the service method return UserResponse and do the mapping there.

src/main/java/jewellery/inventory/controller/UserController.java
```diff
@@ -0,0 +1,66 @@
...
+  public ResponseEntity<UserResponse> getUser(@PathVariable UUID id) {
+    UserResponse userResponse = UserMapper.INSTANCE.toUserResponse(userService.getUser(id));
+    return ResponseEntity.ok(userResponse);
+  }
+
+  @PostMapping
+  public ResponseEntity<UserResponse> createUser(@Valid @RequestBody UserRequest newUser) {
+    UserResponse userResponse =
+        UserMapper.INSTANCE.toUserResponse(
+            userService.createUser(UserMapper.INSTANCE.toUserEntity(newUser)));
```
VladoKat: Same here. Make the service methods take UserRequest and return UserResponse

src/main/java/jewellery/inventory/controller/UserController.java
```diff
@@ -0,0 +1,66 @@
...
+    return ResponseEntity.ok(userResponse);
+  }
+
+  @PutMapping("/{id}")
+  public ResponseEntity<UserResponse> updateUser(
+      @PathVariable UUID id, @Valid @RequestBody UserRequest userRequest) {
+    User userToUpdate = UserMapper.INSTANCE.toUserEntity(userRequest);
+    userToUpdate.setId(id);
+    UserResponse userResponse =
+        UserMapper.INSTANCE.toUserResponse(userService.updateUser(userToUpdate));
```
VladoKat: Same here, the service should take the arguments from the controller as is. The controller shouldn't be responsible for transforming objects.

src/main/java/jewellery/inventory/service/UserService.java
```diff
@@ -0,0 +1,83 @@
...
+import jewellery.inventory.exception.DuplicateNameException;
+import jewellery.inventory.exception.UserNotFoundException;
+import jewellery.inventory.model.User;
+import jewellery.inventory.repository.UserRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class UserService {
+  static final String USER_NOT_FOUND_ERROR_MSG = "User not found with id: ";
```
VladoKat: Should we move the error message in the UserNotFoundException? We can still pass the id as an argument but the core message should be there, I think.

src/main/java/jewellery/inventory/exception/DuplicateNameException.java
```diff
@@ -0,0 +1,7 @@
+package jewellery.inventory.exception;
+
+public class DuplicateNameException extends RuntimeException {
+  public DuplicateNameException() {
+    super("User with that name already exists");
```
VladoKat: How about making this take also an argument that says what name already exists?

src/main/java/jewellery/inventory/exception/DuplicateEmailException.java
```diff
@@ -0,0 +1,8 @@
+package jewellery.inventory.exception;
+
+
+public class DuplicateEmailException extends RuntimeException {
+  public DuplicateEmailException() {
+    super("User with that email already exists");
```
VladoKat: Same as below. Should we pass an argument to the constructor saying what email is duplicated?

src/main/java/jewellery/inventory/mapper/UserMapper.java
```diff
@@ -0,0 +1,21 @@
...
+import jewellery.inventory.dto.UserResponse;
+import jewellery.inventory.model.User;
+import org.mapstruct.Mapper;
+import org.mapstruct.factory.Mappers;
+
+@Mapper(componentModel = "spring")
+public interface UserMapper {
+  UserMapper INSTANCE = Mappers.getMapper(UserMapper.class);
+
+  UserRequest toUserRequest(User user);
```
VladoKat, line 14: I see this method is not used. Should we remove it?

src/test/java/jewellery/inventory/controller/UserControllerTest.java
VladoKat, line 1: I am wondering whether we need this test file at all considering we have created integration tests and the controller itself shouldn't hold much functionality on its own.

src/test/java/jewellery/inventory/helper/UserTestHelper.java
VladoKat, line 1: Let's rename the package name to util and the class to UserTestUtil. I don't have good arguments, honestly. Just prefer util over helper

VladoKat: I found out Baeldung has some arguments for me, though -> https://www.baeldung.com/java-helper-vs-utility-classes#java-helper-vs-utility-classes

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  @Test
+  void willNotUpdateUserWithDuplicateEmail() {
+    // Create first user
+    UserRequest firstUserRequest = createTestUserRequest();
+    ResponseEntity<UserResponse> firstUserResponseEntity =
+        this.testRestTemplate.postForEntity(getBaseUserUrl(), firstUserRequest, UserResponse.class);
+    UserResponse firstUser = firstUserResponseEntity.getBody();
+    assertNotNull(firstUser);
```
VladoKat: Comments are never good to keep in your code. Usually they mislead when trying to explain what the code is doing. In this case, they made me spot a possible method extraction (i.e. sendUserCreateRequest(UserRequest))

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
+
+    // Create second user with different name and email
+    UserRequest secondUserRequest = createDifferentUserRequest();
+    ResponseEntity<UserResponse> secondUserResponseEntity =
+        this.testRestTemplate.postForEntity(
+            getBaseUserUrl(), secondUserRequest, UserResponse.class);
+    UserResponse secondUser = secondUserResponseEntity.getBody();
+    assertNotNull(secondUser);
```
VladoKat: Same as above

src/test/java/jewellery/inventory/helper/UserTestHelper.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.UUID;
+import jewellery.inventory.dto.UserRequest;
+import jewellery.inventory.dto.UserResponse;
+import jewellery.inventory.model.User;
+
+public class UserTestHelper {
+  public static User createTestUser() {
+    User user = new User();
+    user.setName("John");
+    user.setEmail("john@example.com");
```
VladoKat: I noticed some of the values are duplicated within the file. Let's extract them in constants if that's the case.

VladoKat (changes requested):
Minor refactor/extract methods changes. And one a bit bigger to remove the business logic from the controller.

src/test/java/jewellery/inventory/controller/UserControllerTest.java
Alekseeybg: I was wondering the same thing - if I should delete the old Controller tests. And forgot to ask you.

pom.xml
Alekseeybg: Fixed

src/test/java/jewellery/inventory/controller/UserControllerTest.java
VladoKat: I say let's remove the unit tests for the controller.

src/main/java/jewellery/inventory/service/UserService.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/controller/UserController.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/controller/UserController.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/controller/UserController.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/controller/UserControllerTest.java
Alekseeybg: Deleted

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/helper/UserTestHelper.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/DuplicateNameException.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/DuplicateEmailException.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/mapper/UserMapper.java
Alekseeybg: It is used in the tests I believe

VladoKat: It's greyed out in the code and when I search for it there are no mentions:
Screenshot 2023-07-25 at 16 21 54

src/test/java/jewellery/inventory/service/UserServiceTest.java
VladoKat: Are you using it, though? I see the hardcoded strings are still here.

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
```diff
@@ -0,0 +1,290 @@
...
```
VladoKat: These two blocks of code still look pretty similar. Can't we use the sendUserCreateRequest? Also, please, remove the comments. It's not a good practice to use comments unless we absolutely need them.

VladoKat (changes requested):
I saw hardcoded constants are still used in stead of constants. Also, in the CrudIntegrationTest you can replace a lot of the create requests with methods. And please, remove the comments. The code should be self explanatory.

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
Alekseeybg: Just fixed that !

src/main/java/jewellery/inventory/mapper/UserMapper.java
Alekseeybg: image
Something is wrong

Alekseeybg: image

VladoKat (approved):
Looks good! 🎖️

PR #16, opened 2023-07-25
Feature 6/ Expose resource endpoint

VladoKat:
https://trello.com/c/8cwRjWFT/6-be-expose-resource-entity


src/main/java/jewellery/inventory/controller/ResourceController.java
```diff
@@ -0,0 +1,53 @@
...
+@RequestMapping("/resources")
+@RequiredArgsConstructor
+public class ResourceController {
+
+  @Autowired private final ResourceService resourceService;
+
+  @GetMapping
+  public ResponseEntity<List<ResourceResponseDto>> getAllResources() {
+    return ResponseEntity.ok(resourceService.getAllResources());
+  }
```
Alekseeybg: Yesterday while I was refactoring my task I stumbled upon the information that in this case HttpStatus.FOUND is the better choice than OK. This is from my refactor
return ResponseEntity.status(HttpStatus.FOUND).body(userResponseList);

src/main/java/jewellery/inventory/dto/request/resource/GemstoneRequestDto.java
```diff
@@ -0,0 +1,21 @@
...
+
+import lombok.*;
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@Setter
+@ToString(callSuper = true)
+@EqualsAndHashCode(callSuper = true)
+@SuperBuilder
+@NoArgsConstructor
```
Alekseeybg, line 11: @Data contains all these annotations @ToString , @EqualsAndHashCode , @Getter / @Setter and @RequiredArgsConstructor. I was going to suggest to use @Data in all these Dtos but found out that @Data + @SuperBuilder do not work well together so you have no choice but to use separate annotations

src/main/java/jewellery/inventory/service/ResourceService.java
```diff
@@ -0,0 +1,63 @@
...
+
+  public ResourceResponseDto createResource(ResourceRequestDto resourceRequestDto) {
+    return map(resourceRepository.save(map(resourceRequestDto)));
+  }
+
+  public ResourceResponseDto getResourceById(UUID id) {
+    Optional<Resource> resource = resourceRepository.findById(id);
+    if (resource.isEmpty()) {
+      throw new ResourceNotFoundException(id);
+    }
```
Alekseeybg: That if-statement is used in 3 places in this service. Maybe extract it to a method?
We could also use orElseThrow()
```
private Resource getResource(UUID id) {
    return resourceRepository.findById(id)
            .orElseThrow(() -> new ResourceNotFoundException(id));
}
```

src/main/java/jewellery/inventory/service/ResourceService.java
```diff
@@ -0,0 +1,63 @@
...
+import jewellery.inventory.model.resource.Resource;
+import jewellery.inventory.repository.ResourceRepository;
+import lombok.AllArgsConstructor;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+@AllArgsConstructor
+public class ResourceService {
+  @Autowired private ResourceRepository resourceRepository;
```
Alekseeybg: Autowiring through constructors: You are using @AllArgsConstructor to autowire your dependencies. It's recommended to explicitly define the dependencies that your service needs through the constructor. This improves readability and makes it easier to understand what dependencies your class requires. I was using @Autowired in my code as well but refactored it with constructor

src/main/java/jewellery/inventory/service/ResourceService.java
```diff
@@ -0,0 +1,63 @@
...
+
+  public ResourceResponseDto updateResource(UUID id, ResourceRequestDto resourceRequestDto) {
+    Optional<Resource> findResource = resourceRepository.findById(id);
+    if (findResource.isEmpty()) {
+      throw new ResourceNotFoundException(id);
+    }
+    Resource toUpdate = ResourceMapper.toResourceEntity(resourceRequestDto);
+    toUpdate.setId(id);
+    return map(resourceRepository.save(toUpdate));
+  }
```
Alekseeybg: You can make use of the existing entity and set the new values to the existing entity. This will ensure that you're only changing what needs to be changed and keep the rest as it is. Also, it's important to do null checks to avoid NullPointerException:
```
 Resource resource = resourceRepository.findById(id)
      .orElseThrow(() -> new ResourceNotFoundException(id));

    if (resourceRequestDto.getName() != null) {
        resource.setName(resourceRequestDto.getName());
    }
    // similarly for other fields in the ResourceRequestDto

    return map(resourceRepository.save(resource));
```

src/main/java/jewellery/inventory/controller/ResourceController.java
```diff
@@ -0,0 +1,53 @@
...
+import lombok.RequiredArgsConstructor;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@RequestMapping("/resources")
+@RequiredArgsConstructor
+public class ResourceController {
```
Alekseeybg: Maybe we should put
@CrossOrigin(origins = "${cors.origins}") 
I saw you added it to the home controller but not here

src/main/java/jewellery/inventory/mapper/ResourceMapper.java
```diff
@@ -0,0 +1,43 @@
+package jewellery.inventory.mapper;
+
+import jewellery.inventory.dto.request.resource.*;
+import jewellery.inventory.dto.response.resource.*;
+import jewellery.inventory.exception.MappingException;
+import jewellery.inventory.model.resource.*;
+
+public class ResourceMapper {
```
Alekseeybg, line 8: I think instead of class we should be using interface in this case because of the mapstruct library. It uses the interface so it can generate ResourceMapperImpl class with the necessary methods. I am not sure why it works with class also 🗡️

Alekseeybg (changes requested):

src/main/java/jewellery/inventory/controller/ResourceController.java
VladoKat: What information did you stumble upon? From what I read, FOUND means the resource we are trying to get has been moved to a new location: `The HyperText Transfer Protocol (HTTP) 302 Found redirect status response code indicates that the resource requested has been temporarily moved to the URL given by the [Location](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location) header.` https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302#

src/main/java/jewellery/inventory/dto/request/resource/GemstoneRequestDto.java
VladoKat: Yes, in this case we are not using the @ RequiredArgsConstructor. That's why it doesn't make sense to use @ Data

src/main/java/jewellery/inventory/controller/ResourceController.java
Alekseeybg: You are correct. I guess the chatbot (GPT 4) that i am using to consult on some matters was hallucinating again. We should use OK status. And I should fix this in my code as well.

src/main/java/jewellery/inventory/service/ResourceService.java
VladoKat: In a PUT request we are supposed to send all of the fields of the object we are targeting even if the fields are not updated. This means we don't need to do the update field by field. If a field is null in the body, it's just going to be updated to null in the DB. What you are describing would be a PATCH request -> https://www.geeksforgeeks.org/difference-between-put-and-patch-request/

src/main/java/jewellery/inventory/mapper/ResourceMapper.java
VladoKat: This is what ResourceMapperImpl looks like when we define all the default methods: 
Screenshot 2023-07-26 at 12 56 20

This is the ResourceMapper interface: 
Screenshot 2023-07-26 at 13 05 37



It doesn't make any difference whether we define our own class with the implementation or we let it generate it for us in this case because it takes the default implementation and doesn't add anything else.

Alekseeybg: Did you use maven clean install? The code in the Impl class generates after clean install. This is my UserMapperImpl
image

src/main/java/jewellery/inventory/service/ResourceService.java
Alekseeybg: You are correct. To be honest I was not sure if my comment was appropriate for this case

src/main/java/jewellery/inventory/mapper/ResourceMapper.java
VladoKat: This is because you have empty methods that you let MapStruct library generate for you. In the case of ResourceMapper, it doesn't have anything to generate.

Alekseeybg (approved):
Looks good!

PR #17, opened 2023-07-26
[Feature-8-expose-user-endpoint] Refactored the controller and service

Alekseeybg:
Just a quick refactor to make the controller and service as perfect as they could be!


VladoKat (approved):
Looks good!

PR #18, opened 2023-07-26
Feature 8 expose user endpoint

Alekseeybg:


src/main/java/jewellery/inventory/service/UserService.java
```diff
@@ -31,23 +29,18 @@ public UserResponse getUser(UUID id) {
...
+    validateUserDetails(userToCreate);
+    return UserMapper.INSTANCE.toUserResponse(userRepository.save(userToCreate));
   }
 
   public UserResponse updateUser(UserRequest userRequest, UUID id) {
-    User existingUser =
-        userRepository.findById(id).orElseThrow(() -> new UserNotFoundException(id));
+    if (!userRepository.existsById(id)) {
+      throw new UserNotFoundException(id);
+    }
```
VladoKat, line 40: Why did you need to change this? I am curious

VladoKat (commented):
I'm happy to approve. Just curious about the change with existsById instead of findById

src/main/java/jewellery/inventory/service/UserService.java
Alekseeybg: It is supposed to be more performant. With this we are just checking if the user exists while with the old version we checked and had the whole user 'extracted' which was unnecessary

VladoKat (approved):
Looks good

PR #19, opened 2023-07-26
Merge User Endpoint From Test into Main

Alekseeybg:


VladoKat (approved):
Looks good and tested

PR #22, opened 2023-08-04
Feature 12 resource availability

Alekseeybg:


src/main/java/jewellery/inventory/controller/ResourceAvailabilityController.java
```diff
@@ -0,0 +1,60 @@
...
+@RestController
+@RequestMapping("/resources/availability")
+@CrossOrigin(origins = "${cors.origins}")
+@RequiredArgsConstructor
+public class ResourceAvailabilityController {
+  private final ResourceAvailabilityService resourceAvailabilityService;
+
+  @ResponseStatus(HttpStatus.CREATED)
+  @PostMapping
+  public UserResponseDto addResourceToUser(
```
VladoKat: Let's use the ResourceInUserResponseDto instead of the user one.

src/main/java/jewellery/inventory/dto/request/resource/ResourceInUserRequestDto.java
```diff
@@ -0,0 +1,21 @@
...
+@Data
+public class ResourceInUserRequestDto {
+  private static final String QUANTITY_MIN_VALUE_MSG = "Quantity should not be less than 1.";
+  private static final String QUANTITY_DECIMAL_PLACES_MSG =
+      "Quantity should not have more than 2 decimal places.";
+
+  @NotNull private UUID userId;
+  @NotNull private UUID resourceId;
+
+  @Min(value = 1, message = QUANTITY_MIN_VALUE_MSG)
```
VladoKat: quantity could be less than 1. For example, we could want to add 0.5g of gold.

src/main/java/jewellery/inventory/dto/response/UserResponseDto.java
```diff
@@ -1,11 +1,14 @@
...
 import java.util.UUID;
+import jewellery.inventory.dto.response.resource.ResourceInUserResponseDto;
 import lombok.Data;
 
 @Data
 public class UserResponseDto {
   private UUID id;
   private String name;
   private String email;
+  List<ResourceInUserResponseDto> resources;
```
VladoKat: Let's not update the UserResponseDto but modify the ResourceInUserResponseDto to have the UserResponseDto as a field and also a list of ResourceResponseDtos and quantities

src/main/java/jewellery/inventory/dto/response/resource/ResourceInUserResponseDto.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.UUID;
+import lombok.Data;
+
+@Data
+public class ResourceInUserResponseDto {
+  private UUID resourceId;
+  private String ownerName;
+  private String resourceClazz;
+  private double quantity;
```
VladoKat: Can we rework this DTO as mentioned above to have the UserDto as a field and a list of ResourceResponseDtos with quantities. 
 
For the GET /availability/{userId}/{resourceId} we can create a SingleResourceInUserResponseDto that has only one ResourceResponseDto instead of a list. (tbh, I am currently wondering whether we would even need the singular resourceInUser endpoint at all)

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -43,6 +45,18 @@ public ResponseEntity<Object> handleDuplicateException(DuplicateException ex) {
...
+  public ResponseEntity<Object> handleInvalidResourceQuantityException(
+      InvalidResourceQuantityException ex) {
+    return createErrorResponse(HttpStatus.BAD_REQUEST, ex.getMessage());
+  }
+
+  @ExceptionHandler(UserDoesNotOwnResourceException.class)
+  public ResponseEntity<Object> handleUserDoesNotOwnResourceException(
+      UserDoesNotOwnResourceException ex) {
+    return createErrorResponse(HttpStatus.BAD_REQUEST, ex.getMessage());
+  }
```
VladoKat: I notice these methods are pretty much the same. If we have the same error handling for multiple exceptions we can group them in a single method according to https://stackoverflow.com/a/33552395/13062206

src/main/java/jewellery/inventory/exception/duplicateException/DuplicateEmailException.java
```diff
@@ -1,4 +1,4 @@
-package jewellery.inventory.exception;
+package jewellery.inventory.exception.duplicateException;
```
VladoKat: package names should all be lower cased or snake_cased -> https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html 
Also, given the classes are already in an `exception` package, it doesn't make much sense to include it in the name of the inner package (duplicateException). I would name it just `jewellery.inventory.exception.duplicate`

src/main/java/jewellery/inventory/exception/invalidResourceQuantityException/InsufficientResourceQuantityException.java
```diff
@@ -0,0 +1,11 @@
+package jewellery.inventory.exception.invalidResourceQuantityException;
```
VladoKat: same here, rename it to either `invalidresourcequantity` or `invalid_resource_quantity`

src/main/java/jewellery/inventory/exception/notFoundException/NotFoundException.java
```diff
@@ -1,4 +1,4 @@
-package jewellery.inventory.exception;
+package jewellery.inventory.exception.notFoundException;
```
VladoKat: same here -> `notfound` or `not_found`

src/main/java/jewellery/inventory/model/User.java
```diff
@@ -21,8 +22,14 @@ public class User {
...
-  private List<Product> productsOwned;
+  private List<Product> productsOwned = new ArrayList<>();
+
+  @OneToMany(mappedBy = "owner", cascade = CascadeType.ALL, orphanRemoval = true)
+  private List<ResourceInUser> resourcesOwned = new ArrayList<>();
+
+  public void addResource(ResourceInUser resourceInUser) {
+    resourcesOwned.add(resourceInUser);
+    resourceInUser.setOwner(this);
+  }
```
VladoKat: This method is only used in tests from what I saw. Please, move it to the test package in this case. We shouldn't expose methods for global use if we are only using them in the tests.

src/main/java/jewellery/inventory/service/ResourceAvailabilityService.java
VladoKat, line 1: We are now using a double naming convention (I take the blame for it) -> ResourceAvailability and ResourceInUser. Let's just use ResourceInUser everywhere as I think it's more descriptive of what the object/service/repo is doing. We can just keep the endpoint as `resources/availability`.

src/main/java/jewellery/inventory/service/ResourceAvailabilityService.java
```diff
@@ -0,0 +1,118 @@
...
+    resourceInUser.setQuantity(resourceInUser.getQuantity() + resourceUserDto.getQuantity());
+
+    return UserMapper.INSTANCE.toUserResponse(userRepository.save(user));
+  }
+
+  public double getUserResourceQuantity(UUID userId, UUID resourceId) {
+    User user = findUserById(userId);
+    ResourceInUser resourceInUser =
+        findResourceInUser(user, resourceId)
+            .orElseThrow(() -> new ResourceNotFoundException(resourceId));
```
VladoKat: Here we are throwing a `ResourceNotFoundException`. This is a bit misleading as this will be thrown if the resource is not found in the user's possessions.
Let's create a `ResourceInUserNotFoundException` and use it instead.

src/test/java/jewellery/inventory/helper/UserTestHelper.java
```diff
@@ -56,31 +54,40 @@ public static UserRequestDto createDifferentUserRequest() {
...
+    gemstoneRequestDto.setQuantityType("5");
+    gemstoneRequestDto.setColor("blue");
+    gemstoneRequestDto.setCarat(15.0);
+    gemstoneRequestDto.setCut("round");
+    gemstoneRequestDto.setClarity("VS1");
+    gemstoneRequestDto.setDimensionX(1.0);
+    gemstoneRequestDto.setDimensionY(1.0);
+    gemstoneRequestDto.setDimensionZ(1.0);
+    gemstoneRequestDto.setShape("oval");
+    return gemstoneRequestDto;
```
VladoKat: There is a GemstoneRequestDto creation method in ResourceTestHelper -> getGemstoneRequestDto. Please, use it instead

src/main/java/jewellery/inventory/controller/ResourceAvailabilityController.java
```diff
@@ -0,0 +1,60 @@
...
+      @RequestBody @Valid ResourceInUserRequestDto resourceUserDto) {
+    return resourceAvailabilityService.addResourceToUser(resourceUserDto);
+  }
+
+  @ResponseStatus(HttpStatus.OK)
+  @GetMapping("/{userId}/{resourceId}")
+  public double getUserResourceQuantity(@PathVariable UUID userId, @PathVariable UUID resourceId) {
+    return resourceAvailabilityService.getUserResourceQuantity(userId, resourceId);
+  }
```
VladoKat: After some thought, I don't think we need this endpoint at all. Can you please, remove it and all the related methods/dtos.

src/main/java/jewellery/inventory/service/ResourceAvailabilityService.java
```diff
@@ -0,0 +1,118 @@
...
+
+    double totalQuantity = resourceInUser.getQuantity();
+    double newQuantity = totalQuantity - quantity;
+    if (newQuantity < 0) {
+      throw new InsufficientResourceQuantityException(quantity, totalQuantity);
+    }
+
+    resourceInUser.setQuantity(newQuantity);
```
VladoKat: Let's make it so that if we are removing all of the quantity, then we also remove the resource from the user.

src/test/java/jewellery/inventory/integration/ResourceAvailabilityCrudIntegrationTest.java
```diff
@@ -0,0 +1,422 @@
...
+        this.testRestTemplate.exchange(
+            url, HttpMethod.GET, null, new ParameterizedTypeReference<>() {});
+
+    assertEquals(HttpStatus.OK, response.getStatusCode());
+    assertNotNull(response.getBody());
+    assertTrue(response.getBody().isEmpty());
+  }
+
+  @Test
+  void willReturnNotFoundWhenGettingResourcesFromNonExistingUser() {
```
VladoKat: This seems to be duplicate of `willReturnNotFoundWhenUserDoesNotExist`

src/test/java/jewellery/inventory/integration/ResourceAvailabilityCrudIntegrationTest.java
```diff
@@ -0,0 +1,422 @@
...
+    String url = getResourceAvailabilityUrl(createdUser.getId(), nonExistentResourceId) + "/" + 1;
+
+    ResponseEntity<String> response =
+        testRestTemplate.exchange(url, HttpMethod.DELETE, HttpEntity.EMPTY, String.class);
+
+    assertEquals(HttpStatus.NOT_FOUND, response.getStatusCode());
+  }
+
+  @Test
+  void willThrowUserNotExistsExceptionWhenUserDoesNotExist() {
```
VladoKat: This method doesn't indicate it's trying to delete from non existing user. It looks like other methods (GET, PUT) that are dealing with non existing user.

src/test/java/jewellery/inventory/integration/ResourceAvailabilityCrudIntegrationTest.java
VladoKat, line 1: We should think of some naming convention for the tests to indicate what is the difference between them. We are testing a lot of negative scenarios that are very similar to one another - e.g. 1. removing a resource from a user fails for reason x, 2. removing quantity of a resource from a user fails for reason x, 3.fetching resources from a user fails for reason x. All of the mentioned are failing for the same reason but we are testing them for all of the different endpoints. If it were me, I wouldn't have tested all of them but I think I am fine having all of them for now. Right now it's difficult to understand what the methods are trying to do, though. A lot of them are just telling us the outcome of the test, not what caused it. 

TLDR: let's try to think of a naming convention for the tests:
See if this works:
`<action>ResourcesInUser<Sucess/Fail>When<reason>`
examples:
fetchResourcesInUserSuccessfully
fetchResourcesInUserFailsWhenUserNotFound
removeResourcesInUserFailsWhenResourceInUserNotFound

src/test/java/jewellery/inventory/unit/service/ResourceAvailabilityServiceTest.java
```diff
@@ -0,0 +1,378 @@
...
+    verify(userRepository, never()).save(any(User.class));
+  }
+
+  @Test
+  @DisplayName("Should throw ResourceNotFoundException when resource does not exist")
+  void addResourceToUserWhenResourceDoesNotExistThenThrowException() {
+    ResourceInUserRequestDto resourceUserDto =
+        createResourceInUserRequestDto(userId, resourceId, 10);
+
+    when(userRepository.findById(userId)).thenReturn(Optional.of(new User())); // changed this line
```
VladoKat: comment doesn't add any value. please, remove it. if you think there is any template-like/copy-paste between methods, please extract them in methods and use them.

src/test/java/jewellery/inventory/unit/service/ResourceAvailabilityServiceTest.java
VladoKat, line 1: A lot of these tests are covered in the integration tests I think. We should think of a testing strategy because it doesn't make sense to test the same behaviour twice.

VladoKat (changes requested):
There are some changes that need to be made. They are described in the comments. 
Our tests are way too long now. This will make them unmanageable at some point. I will research a bit to see what testing strategy is best to implement but I think the rule of thumb should be:
Have integration tests for the scenarios we think are going to be most commonly used. This may include some negative scenarios but not all of them.
The corner cases and specific error cases scenarios should be tested in service tests.

Let me double check the rules above and we may discuss after how to change (if at all) our testing strategy.

src/main/java/jewellery/inventory/dto/response/resource/ResourceInUserResponseDto.java
VladoKat: Just to make sure we avoid confusion, we are not going to have a SingleResourceInUserResponseDto as it is mentioned in a comment below.

src/main/java/jewellery/inventory/dto/request/resource/ResourceInUserRequestDto.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/notFoundException/NotFoundException.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/invalidResourceQuantityException/InsufficientResourceQuantityException.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/duplicateException/DuplicateEmailException.java
Alekseeybg: While I was renaming them I felt how wrong I did it. Was thinking about asking about it or research it but never managed to do that.

src/test/java/jewellery/inventory/helper/UserTestHelper.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/integration/ResourceAvailabilityCrudIntegrationTest.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/controller/ResourceAvailabilityController.java
Alekseeybg: Removed

src/main/java/jewellery/inventory/service/ResourceAvailabilityService.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/controller/ResourceAvailabilityController.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/dto/response/UserResponseDto.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/dto/response/resource/ResourceInUserResponseDto.java
Alekseeybg: Done

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/service/ResourceAvailabilityService.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/unit/service/ResourceAvailabilityServiceTest.java
Alekseeybg: Removed

src/main/java/jewellery/inventory/model/User.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/integration/ResourceAvailabilityCrudIntegrationTest.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/integration/ResourceAvailabilityCrudIntegrationTest.java
Alekseeybg: Fixed all integration test names as suggested. Both for ResourceInUser and User integration tests

src/main/java/jewellery/inventory/service/ResourceAvailabilityService.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -45,15 +45,9 @@ public ResponseEntity<Object> handleDuplicateException(DuplicateException ex) {
...
-  @ExceptionHandler(InvalidResourceQuantityException.class)
-  public ResponseEntity<Object> handleInvalidResourceQuantityException(
-      InvalidResourceQuantityException ex) {
-    return createErrorResponse(HttpStatus.BAD_REQUEST, ex.getMessage());
-  }
-
-  @ExceptionHandler(UserDoesNotOwnResourceException.class)
-  public ResponseEntity<Object> handleUserDoesNotOwnResourceException(
-      UserDoesNotOwnResourceException ex) {
+  @ExceptionHandler({InvalidResourceQuantityException.class, UserDoesNotOwnResourceException.class})
```
VladoKat: Is `UserDoesNotOwnResourceException` a bad data exception? I assume it's thrown when we are trying to update/delete ResourceInUser that the user doesn't posses. In this case, shouldn't it be a not found?

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -45,15 +45,9 @@ public ResponseEntity<Object> handleDuplicateException(DuplicateException ex) {
...
+  public ResponseEntity<Object> handleResourceExceptions(
```
VladoKat: The name suggests we are handling only resourceExceptions. Let's name it a bit more generic as it is handling all exceptions resulting in BAD_DATA response code: handleBadDataExceptions maybe?

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
VladoKat: Should we include the DuplicateException here as well? I see the `handleDuplicateException` function does the same.

src/main/java/jewellery/inventory/exception/not_found/ResourceInUserNotFoundException.java
```diff
@@ -0,0 +1,9 @@
+package jewellery.inventory.exception.not_found;
+
+import java.util.UUID;
+
+public class ResourceInUserNotFoundException extends NotFoundException {
+    public ResourceInUserNotFoundException(UUID id) {
+        super("The resource with id " + id + " is not owned by this user");
```
VladoKat: Could we also include the user UUID in this exception?

src/main/java/jewellery/inventory/mapper/UserMapper.java
```diff
@@ -26,19 +27,23 @@ public interface UserMapper {
...
-        resourcesDtoList.add(toResourceInUserResponseDto(resourceInUser));
-      }
-      userResponseDto.setResources(resourcesDtoList);
+  default void enrichResourceInUserResponseDto(
+      ResourceInUser resourceInUser,
+      @MappingTarget ResourceInUserResponseDto resourceInUserResponseDto) {
+    if (resourceInUser.getResource() != null) {
+      ResourceResponseDto resourceResponseDto =
+          ResourceMapper.toResourceResponse(resourceInUser.getResource());
+      resourceInUserResponseDto.setResources(Collections.singletonList(resourceResponseDto));
```
VladoKat: I might be getting this wrong but it seems to me the ResourceInUserResponseDto will always have just the one resource inside it. It could have multiple

src/main/java/jewellery/inventory/service/ResourceInUserService.java
```diff
@@ -73,6 +68,9 @@ public void removeQuantityFromResource(UUID userId, UUID resourceId, double quan
     }
 
     resourceInUser.setQuantity(newQuantity);
+    if (newQuantity == 0) {
```
VladoKat: Have you checked for newQuantity = 0.00? I remember there was some weird behaviour comparing double to 0.

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
```diff
@@ -75,28 +78,31 @@ void setup() {
...
 
     ResourceInUserRequestDto requestDto =
         createResourceInUserRequestDto(
             createdUser.getId(), createdResource.getId(), RESOURCE_QUANTITY);
 
-    ResponseEntity<UserResponseDto> response =
-        testRestTemplate.postForEntity(
-            getBaseResourceAvailabilityUrl(), requestDto, UserResponseDto.class);
+    System.out.println(createdUser.getId());
+    System.out.println(createdResource.getId());
```
VladoKat: let's not leave system.out in the code pls

src/main/java/jewellery/inventory/dto/response/resource/ResourceInUserResponseDto.java
```diff
@@ -1,12 +1,12 @@
...
+import jewellery.inventory.dto.response.UserResponseDto;
 import lombok.Data;
 
 @Data
 public class ResourceInUserResponseDto {
-  private UUID resourceId;
-  private String ownerName;
-  private String resourceClazz;
+  private UserResponseDto owner;
+  List<ResourceResponseDto> resources;
```
VladoKat: The list should be a list of (resource, quantity) <=> List<Pair<ResourceResponseDto, Double>>

VladoKat (changes requested):
Please, take a look at the comments. There is one major thing that needs to be changed - the ResourceInUserResponseDto. It could be renamed to Resource**s**InUserResponseDto.

Another thing that is not mentioned but we might want to do is to make the ResourceInUser<smth> to be outside the resource packages in the model and dto packages. It feels a bit out of place there as the other are direct descendants of Resource.

src/main/java/jewellery/inventory/dto/response/resource/ResourceInUserResponseDto.java
VladoKat: ResourceInUserRequest should remain as it is.

src/main/java/jewellery/inventory/exception/not_found/ResourceInUserNotFoundException.java
VladoKat: What is the difference between this and `UserDoesNotOwnResourceException`?

src/main/java/jewellery/inventory/service/ResourceInUserService.java
```diff
@@ -0,0 +1,116 @@
...
+        findResourceInUser(user, resourceUserDto.getResourceId())
+            .orElseGet(() -> createAndAddNewResourceInUser(user, resource, 0));
+
+    resourceInUser.setQuantity(resourceInUser.getQuantity() + resourceUserDto.getQuantity());
+
+    return UserMapper.INSTANCE.toResourceInUserResponseDto(
+        resourceInUserRepository.save(resourceInUser));
+  }
+
+  public List<ResourceInUserResponseDto> getAllResourcesFromUser(UUID userId) {
```
VladoKat: After the ResourcesInUserResponseDto, it would make this gets all the ResourceInUser owned by a user and returns a single ResourcesInUserResponseDto

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
```diff
@@ -0,0 +1,386 @@
...
+    return "http://localhost:" + port + "/resources/availability/" + userId + "/" + resourceId;
+  }
+
+  private String getBaseUserUrl() {
+    return "http://localhost:" + port + "/users";
+  }
+
+  private String getBaseResourceUrl() {
+    return "http://localhost:" + port + "/resources";
+  }
```
VladoKat: Lots of hardcoded strings repetitions. Could we extract at least getBaseUrl() to return "http://localhost:" + port

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/dto/response/resource/ResourceInUserResponseDto.java
Alekseeybg: Created a separate DTO to fix the issue. I believe this is the correct approach in this case.

src/main/java/jewellery/inventory/exception/not_found/ResourceInUserNotFoundException.java
Alekseeybg: Removed UserDoesNotOwnResourceException. I really do not know why I basically duplicated it... I was thinking of including the user UUID as well but did not want the error message to be too long. Fixed.

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
Alekseeybg: It was a duplicate of ResourceInUserNotFoundException. Deleted.

src/main/java/jewellery/inventory/dto/response/resource/ResourceInUserResponseDto.java
VladoKat: What is the scenario in which we are going to use this response? The list of resource dtos can't map to a single quantity field.

src/main/java/jewellery/inventory/service/ResourceInUserService.java
Alekseeybg: Fixed.

src/main/java/jewellery/inventory/service/ResourceInUserService.java
Alekseeybg: I did not know about that issue! Fixed.

src/main/java/jewellery/inventory/mapper/UserMapper.java
Alekseeybg: I think you were correct. Not sure if the refactored code fixed this issue though.

src/main/java/jewellery/inventory/dto/response/resource/ResourceInUserResponseDto.java
Alekseeybg: At the time you wrote your comment the new code was not pushed yet. The new DTO is not a response DTO. It is used in the ResourcesInUserResponseDto. The new version is
```@Data
public class ResourcesInUserResponseDto {
  private UserResponseDto owner;
  private List<ResourceQuantityDto> resources;
}
```
And the new DTO looks like this:
```@Data
public class ResourceQuantityDto {
    private ResourceResponseDto resource;
    private double quantity;
}
```
Maybe the naming ResourceQuantityDto  is not the best

VladoKat: yep, this looks better. I don't mind the name. Maybe more appropriate would be ResourceAndQuantityDto

src/main/java/jewellery/inventory/service/ResourceInUserService.java
```diff
@@ -40,15 +43,22 @@ public ResourceInUserResponseDto addResourceToUser(ResourceInUserRequestDto reso
...
+    List<ResourceQuantityDto> resourcesWithQuantities =
+        user.getResourcesOwned().stream()
+            .map(resourcesInUserMapper::toResourceQuantityDto)
+            .collect(Collectors.toList());
+
+    ResourcesInUserResponseDto responseDto = new ResourcesInUserResponseDto();
+    responseDto.setOwner(resourcesInUserMapper.toUserResponse(user));
+    responseDto.setResources(resourcesWithQuantities);
+
+    return responseDto;
```
VladoKat, line 61: Can we make a mapping method in.. (UserMapper?) that maps User (entity) to ResourcesInUserResponseDto?

VladoKat (commented):
One minor comment. Would approve after we discuss it.

src/main/java/jewellery/inventory/service/ResourceInUserService.java
Alekseeybg: Added it to ResourcesInUserMapper

VladoKat:
duplicate of https://github.com/JewelleryManagement/jewellery-management-service/pull/23

PR #32, opened 2023-09-27
Implemented user authentication

Alekseeybg:


pom.xml
```diff
@@ -26,6 +26,35 @@
...
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-api</artifactId>
+            <version>0.11.5</version>
```
VladoKat: Extract jjwt-version in a property as we are using it multiple times.

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
```diff
@@ -9,18 +9,28 @@
...
+      "^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,6}$";
+  private static final String PWD_PATTERN_REGEX =
+      "^(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z])(?=.*[@#$%^&+=])(?=\\S+$).{8,}$";
   private static final String NAME_SIZE_VALIDATION_MSG = "Size must be between 3 and 64";
+  private static final String PWD_SIZE_VALIDATION_MSG = "Size must be at least 8 characters";
   private static final String NAME_PATTERN_VALIDATION_MSG =
       "Name must only contain alphanumeric characters and underscores, and no consecutive underscores";
   private static final String EMAIL_VALIDATION_MSG = "Email must be valid";
+  private static final String PWD_PATTERN_VALIDATION_MSG =
+      "Password must contain at least one digit, one lowercase letter, one uppercase letter, one special character, and be at least 8 characters long";
```
VladoKat: Do we need all these to be constanst? We are only using them once. I'd prefer to keep them hardcoded in this case. Also, I see for the @ NotBlank you haven't created constants. Let's stick to either hardcode everywhere use constants everywhere. I'd go for the first approach if we only have them in one place.

VladoKat: Maybe have a constant just for the very long messages.

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -46,6 +49,11 @@ public ResponseEntity<Object> handleBadDataExceptions(RuntimeException ex) {
     return createErrorResponse(HttpStatus.BAD_REQUEST, ex.getMessage());
   }
 
+  @ExceptionHandler({JwtAuthenticationBaseException.class,  SignatureException.class, AuthenticationException.class})
+  public ResponseEntity<Object> handleExpiredJwtException(AuthenticationException ex) {
```
VladoKat: name this `handleAuthenticationException`?

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -46,6 +49,11 @@ public ResponseEntity<Object> handleBadDataExceptions(RuntimeException ex) {
     return createErrorResponse(HttpStatus.BAD_REQUEST, ex.getMessage());
   }
 
+  @ExceptionHandler({JwtAuthenticationBaseException.class,  SignatureException.class, AuthenticationException.class})
```
VladoKat: I see `JwtAuthenticationBaseException` extends `AuthenticationException`. Can we remove the `JwtAuthenticationBaseException` from here as it should be caught anyway as child of `AuthenticationException`

src/main/java/jewellery/inventory/exception/not_found/UserNotFoundException.java
```diff
@@ -6,4 +6,8 @@ public class UserNotFoundException extends NotFoundException {
   public UserNotFoundException(UUID id) {
     super("User with id " + id + " was not found");
   }
+
+  public UserNotFoundException() {
+    super("User was not found");
+  }
```
VladoKat: Let's not have such a generic message. I see we are using this when we try to find a user by email. Create a constructor that takes an email and add it in the message.

src/main/java/jewellery/inventory/model/User.java
```diff
@@ -20,9 +28,50 @@ public class User {
...
 
+  @Column(unique = true)
+  private String password;
+
+  @Column
+  @Enumerated(EnumType.STRING)
+  private Role role;
+
+  // This field is essential for the application's business logic.
+  // It is not marked as transient despite SonarQube's recommendation.
```
VladoKat: These comments are not needed. I was looking into the issue and it seems we can't easily get rid of it. I think we are going to just mark it as "Won't fix". The comments are obsolete either way.

src/main/java/jewellery/inventory/security/JwtAuthenticationEntryPoint.java
```diff
@@ -0,0 +1,30 @@
...
+      HttpServletRequest request,
+      HttpServletResponse response,
+      AuthenticationException authException) {
+    resolver.resolveException(request, response, null, authException);
+  }
+
+  public void commence(
+      HttpServletRequest request, HttpServletResponse response, Exception exception) {
+    resolver.resolveException(request, response, null, exception);
+  }
```
VladoKat: Why do we have both of these methods? The only place where the Exception one is used is in JwtAuthenticationFilter->clearSecurityContextAndCommenceEntryPoint and that's because for some reason we are expecting Exception as argument and when we call that method in doFilterInternal we are catching AuthenticationException. So if we just change the Exception parameter to AuthenticationException, we shouldn't need the second commence method.

src/main/java/jewellery/inventory/security/JwtTokenService.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  public boolean isTokenValid(String token, UserDetails userDetails) {
+    try {
+      final String name = extractName(token);
+      if (!name.equals(userDetails.getUsername())) {
+        throw new JwtIsNotValidException();
+      }
+    } catch (SignatureException e) {
+      throw new JwtIsNotValidException();
```
VladoKat: Where do we expect the error to be thrown from here? I suggest putting only that part in the try block.

VladoKat: also, let's extract this into a isNameValid(token) method

src/main/java/jewellery/inventory/security/JwtTokenService.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    try {
+      if (isTokenExpired(token)) {
+        throw new JwtExpiredException();
+      }
+    } catch (SignatureException e) {
+      throw new JwtIsNotValidException();
+    }
```
VladoKat: Can we just add the catch SignatureException in the isTokenExpired method and then we will just call isTokenExpired here

src/main/java/jewellery/inventory/security/JwtUtils.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Value("${jwt.secret.key}")
+  private String secretKey;
+
+  public Key getSigningKey() {
+    try {
+      byte[] keyBytes = Decoders.BASE64.decode(secretKey);
+      return Keys.hmacShaKeyFor(keyBytes);
+    } catch (WeakKeyException e) {
+      throw new InvalidSecretKeyException();
```
VladoKat: Let's pass e to the InvalidSecretKeyException constructor and propagate the cause and message:
`public InvalidSecretKeyException(WeakKeyException e) {
    super("Invalid secret key: " + e.getMessage(), e.getCause());
  }`
We should also create a test for this.

src/main/java/jewellery/inventory/security/JwtUtils.java
```diff
@@ -0,0 +1,45 @@
...
+    try {
+      return Jwts.parserBuilder()
+          .setSigningKey(getSigningKey())
+          .build()
+          .parseClaimsJws(token)
+          .getBody();
+    } catch (ExpiredJwtException e) {
+      throw new JwtExpiredException();
+    } catch (SignatureException e) {
+      throw new JwtIsNotValidException();
```
VladoKat: Same here, we should pass the information from the incoming exception to our custom ones.

src/main/java/jewellery/inventory/security/SecurityConfig.java
```diff
@@ -0,0 +1,88 @@
...
+            exceptionHandling ->
+                exceptionHandling.authenticationEntryPoint(jwtAuthenticationEntryPoint))
+        .addFilterBefore(jwtAuthenticationFilter, UsernamePasswordAuthenticationFilter.class);
+
+    AuthorizeHttpRequestsConfigurer<HttpSecurity> authorizeRequests =
+        new AuthorizeHttpRequestsConfigurer<>(applicationContext);
+    authorizeRequests
+        .getRegistry()
+        .requestMatchers(HttpMethod.POST, USERS_ENDPOINT)
+        .permitAll()
```
VladoKat: We should not have the users endpoint allowed for all. Let's go for requiring all functional endpoints to be protected. Let's create a root admin user with hardcoded email and password that would allow us to start interacting with the backend initially.

src/main/java/jewellery/inventory/service/AuthService.java
```diff
@@ -0,0 +1,54 @@
...
+  private final UserRepository repository;
+  private final AuthenticationManager authenticationManager;
+  private final JwtTokenService jwtService;
+
+  public AuthenticationResponseDto authenticate(AuthenticationRequestDto authRequest) {
+    if (!tryAuthenticate(authRequest)) {
+      throw new InvalidCredentialsException();
+    }
+
+    User user = getUserByEmail(authRequest.getEmail()).orElseThrow(UserNotFoundException::new);
```
VladoKat: Add the email to the message of UserNotFoundException

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.crypto.password.PasswordEncoder;
+
+class AuthIntegrationTest extends AuthenticatedIntegrationTestBase {
+
+  @Autowired private UserRepository userRepository;
+  @MockBean private PasswordEncoder passwordEncoder;
```
VladoKat, line 31: Do we need to mock the PasswordEncoder?

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Override
+  @BeforeEach
+  void setup() {
+    userRepository.deleteAll();
+    SecurityContextHolder.clearContext();
+    mockPasswordEncoder();
+    createAndSaveTestUser();
+    when(userDetailsService.loadUserByUsername(anyString())).thenReturn(testUser);
+    setupAuthRequestAndHeaders();
```
VladoKat: From what I understand, we should only be calling this once, and not per each test.

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
```diff
@@ -0,0 +1,127 @@
...
+    assertNotNull(responseBody);
+    assertTrue(responseBody.contains(USER_NAME));
+    assertTrue(responseBody.contains(USER_EMAIL));
+  }
+
+  @Test
+  void testInvalidUrlReturnsUnauthorized() {
+    ResponseEntity<Void> response =
+        testRestTemplate.exchange("/invalid-url", HttpMethod.GET, null, Void.class);
+    assertEquals(HttpStatus.UNAUTHORIZED, response.getStatusCode());
```
VladoKat: I don't think we need this test case. We are supposed to return 401 for every request that is not authorized (except for the post to /auth/token). If a request is authorized but the url is not valid, then we should return 404. Is this not the case currently?

src/test/java/jewellery/inventory/unit/service/security/JwtTokenServiceTest.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @Test
+  void generateTokenWithExtraClaimsReturnsCorrectClaims() {
+    Map<String, Object> extraClaims = createExtraClaims();
+
+    String token = jwtTokenService.generateToken(extraClaims, userDetails);
+
+    Claims claims = parseClaimsFromToken(token);
+    assertThat(claims).containsEntry(CLAIM_KEY, CLAIM_VALUE);
```
VladoKat: From what I saw, we never generate extra claims. What are those claims used for? If we have no use for it, there is no point in having the test or the logic for it.

VladoKat (changes requested):
The logic for validating the jwt token seems more complex than I thought it would be. We have a lot of exception handling that I am not sure when we are supposed to hit. If we have cases in which these scenarios could happen, then I suggest creating tests for  them. If not, we should consider simplifying the code. (I am not saying I know how but would love to dive deeper and try to understand it)

src/main/java/jewellery/inventory/model/User.java
Alekseeybg: I marked it as "Won't fix" (found it where it was yesterday). Will remove the comments

pom.xml
Alekseeybg: Fixed

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
Alekseeybg: Fixed. Left only NAME_PATTERN_VALIDATION_MSG and PWD_PATTERN_VALIDATION_MSG as constants because are too long.

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/exception/not_found/UserNotFoundException.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/model/User.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/security/JwtAuthenticationEntryPoint.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/security/JwtTokenService.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/service/AuthService.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: I think yes. Because we create a testUser and to save it properly in the mock database the password should be encoded using the encoder... If I am missing something let me know.

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: Removed it. The idea came from a colleague of mine who told me what integration tests they write for their project.. And with spring security I do not know how to make it so it returns 404 Not found instead of unauthorized for not existing endpoints... I think the default behavior is to return always unauthorized. I just tested - tried to hit not existing endpoint and provided valid token in the headers - it returned 401 unauthorized.

src/test/java/jewellery/inventory/unit/service/security/JwtTokenServiceTest.java
Alekseeybg: Removed the logic/test.

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
VladoKat: but with the other integration tests the database isn't mocked. Are we mocking it here?

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
VladoKat: hm, that's interesting. I wonder why...

src/main/java/jewellery/inventory/security/JwtTokenService.java
Alekseeybg: It did not work as intended for some reason. So the catch is in the extractAllClaims method in JwtUtils class. Only then the correct message is shown. I refactored that part of the code

src/main/java/jewellery/inventory/security/SecurityConfig.java
Alekseeybg: Made the needed adjustments to the code and added the sql script to add a root user with admin access... I added role USER and when a user is created it defaults to this role.

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: Refactored.

src/main/java/jewellery/inventory/security/JwtUtils.java
Alekseeybg: Fixed. Will add test/s soon.

src/main/java/jewellery/inventory/security/JwtUtils.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: I think for this we need valid (existing) user with valid credentials provided (existing email and existing encoded password) so it generates token. Will check if the process could be simplified

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: Probably because we override the default behavior and tell it to return 401 instead of 403 🗡️

src/main/java/jewellery/inventory/security/SecurityConfig.java
```diff
@@ -0,0 +1,76 @@
...
+            session -> session.sessionCreationPolicy(SessionCreationPolicy.STATELESS))
+        .cors(Customizer.withDefaults())
+        .exceptionHandling(
+            exceptionHandling ->
+                exceptionHandling.authenticationEntryPoint(jwtAuthenticationEntryPoint))
+        .addFilterBefore(jwtAuthenticationFilter, UsernamePasswordAuthenticationFilter.class);
+
+    http.authorizeHttpRequests(
+        authorize ->
+            authorize.requestMatchers(AUTH_ENDPOINT).permitAll().anyRequest().hasRole("ADMIN"));
```
VladoKat: Let's use the Role enum's ADMIN.name() for this one.

src/test/java/jewellery/inventory/unit/service/security/AuthServiceTest.java
```diff
@@ -0,0 +1,112 @@
...
+  @BeforeEach
+  public void setUp() {
+    authRequest = new AuthenticationRequestDto();
+    authRequest.setEmail(USER_EMAIL);
+    authRequest.setPassword(USER_PASSWORD);
+
+    user = new User();
+    user.setId(UUID.randomUUID());
+    user.setEmail(USER_EMAIL);
+    user.setPassword(USER_PASSWORD);
```
VladoKat: Isn't this the same as UserTestHelper.createTestUser() ? Let's use it.

src/test/java/jewellery/inventory/unit/service/security/JwtTokenServiceTest.java
```diff
@@ -230,4 +233,19 @@ private void setupDefaultExtractAllClaimsBehavior() {
...
     lenient().doReturn(claims).when(jwtUtils).extractAllClaims(anyString());
   }
+
+  private static Stream<String> invalidEmailProvider() {
+    return Stream.of(null, "", " ");
+  }
+
+  private String generateExpiredToken() {
+    Date expiredDate =
+        new Date(System.currentTimeMillis() - (1000 * 60 * 60 * 24)); // 1 day in the past
```
VladoKat: Please, remove the comment. You can, instead, extract the complex calculation (1000 * 60 * 60 * 24) in a variable named dayInMillis.

VladoKat (changes requested):

src/test/java/jewellery/inventory/unit/service/security/AuthServiceTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Test
+  void authenticateWithNonExistentUserThrowsUserNotFoundException() {
+    when(userRepository.findByEmail(authRequest.getEmail())).thenReturn(Optional.empty());
+
+    assertThrows(UserNotFoundException.class, () -> authService.login(authRequest));
+  }
+
+  @Test
+  void verifyMethodsCalledWithExpectedParameters() {
```
VladoKat: This method name is too generic. The setup is the same as `authenticateWithValidCredentialsReturnsToken.` Should we move the verification methods to `authenticateWithValidCredentialsReturnsToken`

src/test/java/jewellery/inventory/unit/service/security/AuthServiceTest.java
```diff
@@ -0,0 +1,112 @@
...
+    UserResponseDto userResponseDto = new UserResponseDto();
+    userResponseDto.setEmail(USER_EMAIL);
+
+    when(userMapper.toUserResponse(user)).thenReturn(userResponseDto);
+
+    UserAuthDetailsDto response = authService.login(authRequest);
+
+    assertEquals(userResponseDto, response.getUser());
+    assertEquals(TEST_TOKEN, response.getToken());
+  }
```
VladoKat: Same question with this method. Is the setup any different than `authenticateWithValidCredentialsReturnsToken`? What are we expecting differently?

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+    ResponseEntity<UserAuthDetailsDto> response =
+        testRestTemplate.exchange(
+            getBaseAuthUrl(), HttpMethod.POST, requestEntity, UserAuthDetailsDto.class);
+
+    assertEquals(HttpStatus.CREATED, response.getStatusCode());
+  }
+
+  @Test
+  void testUnauthorizedAccessWithInvalidToken() {
```
VladoKat: Let's stick to the test name pattern (e.g. fetchUsersWithInvalidTokenWillReturnUnauthorized)

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+    String invalidToken = "invalid_token";
+    addTokenToHeaders(invalidToken);
+    HttpEntity<Void> requestEntity = new HttpEntity<>(headers);
+    ResponseEntity<Void> response =
+        testRestTemplate.exchange(getBaseUserUrl(), HttpMethod.GET, requestEntity, Void.class);
```
VladoKat, line 61: This is getBaseUrl is a bit misleading. I thought it's just baseUrl. It has users after it. Please change it to reflect what we are actually calling

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+    assertEquals(HttpStatus.UNAUTHORIZED, response.getStatusCode());
+  }
+
+  @Test
+  void testAccessToProtectedEndpointWithValidToken() {
```
VladoKat: Same here - fetchUsersWithValidTokenWillSucceed

VladoKat: BTW, is this different than the successful case in the Users integration test in this case?

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+    String responseBody = response.getBody();
+
+    assertNotNull(responseBody);
+    assertTrue(responseBody.contains(USER_NAME));
+    assertTrue(responseBody.contains(USER_EMAIL));
+  }
+
+  @Test
+  void requestWithNullEmailShouldReturnBadRequest() {
```
VladoKat: loginWithNullEmailCredentialReturnsBadRequest

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+        testRestTemplate.exchange(getBaseAuthUrl(), HttpMethod.POST, requestEntity, String.class);
+
+    assertEquals(HttpStatus.BAD_REQUEST, response.getStatusCode());
+    assertNotNull(response.getBody());
+    assertTrue(response.getBody().contains("email"));
+    assertTrue(response.getBody().contains("Email must not be blank, empty or null"));
+  }
+
+  @Test
+  void requestWithNullPasswordShouldReturnBadRequest() {
```
VladoKat: loginWithNullPasswordReturnsBadRequest

VladoKat (changes requested):
Apart from the comments, here, please change passwords to be char arrays instead of Strings. Let me know if it turns out to be too much of a hassle.

src/test/java/jewellery/inventory/unit/service/security/AuthServiceTest.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/unit/service/security/JwtTokenServiceTest.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/unit/service/security/AuthServiceTest.java
Alekseeybg: The first test was written before the refactor of the response we are returning. At first we returned just the token so we tested just for that. The second test was written after the refactor and in this case we check if the user data returned is correct as well. But you are correct it should be only one test. Fixed by merging.

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: it is getBaseUserUrl() but it is hard to see there is User in there :)

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: You are correct. Did not realize it will duplicate old test. Removed it.

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/unit/service/security/AuthServiceTest.java
Alekseeybg: Merged with authenticateWithValidCredentialsReturnsToken

src/test/java/jewellery/inventory/unit/service/security/JwtTokenServiceTest.java
VladoKat: I see you ended up not extracting `dayInMillis = (1000 * 60 * 60 * 24)`. Any reason why not?

src/test/java/jewellery/inventory/integration/AuthIntegrationTest.java
VladoKat: Alt Text

VladoKat (approved):
Looks good!

PR #33, opened 2023-10-10
Feature 15 implement product

ignatIgnatov:
implemented Product


src/main/java/jewellery/inventory/dto/response/ProductResponseDto.java
```diff
@@ -0,0 +1,28 @@
...
+    private UUID id;
+    private String name;
+    private List<String> authors;
+
+    private UserResponseDto owner;
+
+    private List<ResourceInProductResponseDto> resourcesContent;
+    private List<ProductResponseDto> productsContent;
+
+    private UUID contentId;
```
VladoKat: should we rename this to `contentOf`? Right now this tells me this is the id of the content of this product.

src/main/java/jewellery/inventory/dto/response/resource/ResourceInProductResponseDto.java
VladoKat, line 1: This DTO is the same as `ResourceQuantityDto`. I suggest Renaming `ResourceQuantityDto` to `ResourceQuantityResponseDto` and moving it to response folder. And deleting this one. Then to keep things consistent we might also rename `ResourceInProductRequestDto` to `ResourceQuantityRequestDto`.

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -46,6 +49,21 @@ public ResponseEntity<Object> handleBadDataExceptions(RuntimeException ex) {
     return createErrorResponse(HttpStatus.BAD_REQUEST, ex.getMessage());
   }
 
+  @ExceptionHandler({ResourcesInProductNotFoundException.class})
+  public ResponseEntity<Object> handleProductCantBeCreatedException(NotFoundException ex) {
+    return createErrorResponse(HttpStatus.NOT_FOUND, ex.getMessage());
```
VladoKat: This method is the same as `handleNotFound` method. `ResourcesInProductNotFoundException` extends NotFoundException, so it will be automatically handled by the `handleNotFoundMethod`

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -46,6 +49,21 @@ public ResponseEntity<Object> handleBadDataExceptions(RuntimeException ex) {
...
+
+  @ExceptionHandler({ProductContainsException.class})
+  public ResponseEntity<Object> handleProductContainsException(RuntimeException ex) {
+    return createErrorResponse(HttpStatus.CONFLICT, ex.getMessage());
+  }
+
+  @ExceptionHandler({ProductIsSoldException.class})
+  public ResponseEntity<Object> handleProductIsSoldException(RuntimeException ex) {
+    return createErrorResponse(HttpStatus.CONFLICT, ex.getMessage());
+  }
```
VladoKat: Both of these methods are doing the same thing. You can combine them in a single `handleEntityConstraintConflict` method by providing `@ExceptionHandler({ProductContainsException.class, ProductIsSoldException.class})`.

I would rename the `ProductContainsException` as it's not clear what kind of a problem this is. I would rename it to smth like `ProductIsContentException`.

src/main/java/jewellery/inventory/exception/invalid_resource_quantity/NegativeResourceQuantityException.java
```diff
@@ -2,6 +2,6 @@
 
 public class NegativeResourceQuantityException extends InvalidResourceQuantityException {
   public NegativeResourceQuantityException(double quantity) {
-    super("Resource Quantity you entered is negative: " + quantity);
+    super("Resource Quantity you entered is negative: -" + quantity);
```
VladoKat: I am a bit surprised you need to add this. If quantity is negative, shouldn't it be shown as it is and it would have '-'?

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -46,6 +49,21 @@ public ResponseEntity<Object> handleBadDataExceptions(RuntimeException ex) {
...
   }
 
+  @ExceptionHandler({ResourcesInProductNotFoundException.class})
+  public ResponseEntity<Object> handleProductCantBeCreatedException(NotFoundException ex) {
+    return createErrorResponse(HttpStatus.NOT_FOUND, ex.getMessage());
+  }
+
+  @ExceptionHandler({ProductContainsException.class})
+  public ResponseEntity<Object> handleProductContainsException(RuntimeException ex) {
+    return createErrorResponse(HttpStatus.CONFLICT, ex.getMessage());
```
VladoKat, line 58: Good use of HttpStatus.CONFLICT! I haven't seen this one.

src/main/java/jewellery/inventory/exception/not_found/ResourceNotFoundException.java
```diff
@@ -6,4 +6,8 @@ public class ResourceNotFoundException extends NotFoundException {
     public ResourceNotFoundException(UUID id) {
         super("The resource with id " + id + " is not found");
     }
+
+    public ResourceNotFoundException(String name) {
+        super("The resource with name " + name + " is not found");
+    }
```
VladoKat: What is the name you are suggesting here? Resources don't have names

src/main/java/jewellery/inventory/exception/not_found/ResourcesInProductNotFoundException.java
```diff
@@ -0,0 +1,9 @@
+package jewellery.inventory.exception.not_found;
+
+import jewellery.inventory.exception.not_found.NotFoundException;
+
+public class ResourcesInProductNotFoundException extends NotFoundException {
```
VladoKat: I wouldn't name this a NotFound exception. It suggests we can't find some resources in the database. If I get this correctly based on the message, this should be a ... `ResourcesInProductMissingException`. The difference is subtle but NotFound implies 404 code, whereas we are validating that the product we are creating should have some resources from which it is created, which is more in the realm of 400 Bad Request -> user needs to provide some resources to create the product out of.

src/main/java/jewellery/inventory/exception/product/ProductContainsException.java
```diff
@@ -0,0 +1,9 @@
+package jewellery.inventory.exception.product;
+
+import java.util.UUID;
+
+public class ProductContainsException extends RuntimeException {
```
VladoKat: As mentioned in a comment above, let's rename this to `ProductIsContentException` or something similar.

src/main/java/jewellery/inventory/exception/product/ProductContainsException.java
```diff
@@ -0,0 +1,9 @@
...
+    public ProductContainsException(UUID id) {
+        super("The product with id " + id + " contains in another product!");
```
VladoKat: Let's change the message to `The product with id " + id + " is content of another product!`

src/main/java/jewellery/inventory/mapper/ProductMapper.java
```diff
@@ -0,0 +1,11 @@
+package jewellery.inventory.mapper;
+
+import jewellery.inventory.dto.response.ProductResponseDto;
+import jewellery.inventory.model.Product;
+import org.mapstruct.Mapper;
+
+@Mapper(componentModel = "spring")
+public interface ProductMapper {
+
+    ProductResponseDto toProductResponse(Product product);
```
VladoKat: I am pretty certain this method wouldn't map the `ResourceResponseDto`s. It will just map the fields of the parent Resource. We should implement a custom logic for mappers that are mapping resources unfortunately.

src/main/java/jewellery/inventory/mapper/ResourceInProductMapper.java
VladoKat, line 1: I believe the same as for the above file would go here.

src/main/java/jewellery/inventory/model/Product.java
```diff
@@ -1,30 +1,47 @@
...
+    @ManyToOne
+    private User owner;
+
+    @OneToMany(mappedBy = "product", cascade = CascadeType.ALL)
+    private List<ResourceInProduct> resourcesContent;
+
+    @JsonIgnore
+    @ManyToOne
+    @JoinColumn(name = "content_id")
+    private Product content;
```
VladoKat: let's name this `contentOf`. Then it would make sense to rename the JoinColumn as well to content_of?

src/main/java/jewellery/inventory/model/ResourceInUser.java
```diff
@@ -19,8 +20,10 @@
 public class ResourceInUser {
   @Id @GeneratedValue private UUID id;
 
+  @JsonIgnore
   @ManyToOne private User owner;
 
+  @JsonIgnore
```
VladoKat: Why do you need to exclude these when serializing?

src/main/java/jewellery/inventory/model/resource/ResourceInProduct.java
```diff
@@ -15,9 +16,13 @@
...
+  @JsonIgnore
+  @ManyToOne
+  private Resource resource;
 
   private double quantity;
 
-  @ManyToOne private Product product;
+  @JsonIgnore
+  @ManyToOne
+  private Product product;
```
VladoKat: Same question here. Why do we need to `@JsonIgnore` these fields?

src/main/java/jewellery/inventory/repository/ProductRepository.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+import java.util.UUID;
+
+@Repository
+public interface ProductRepository extends JpaRepository<Product, UUID> {
+
+    Optional<Product> findByName(String name);
```
VladoKat: This method is not used anywhere, I think. Please, remove it.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,260 @@
...
+            throw new ResourceInUserNotFoundException(user.getId());
+        }
+
+        List<ResourceInProduct> resourcesInProducts = new ArrayList<>();
+
+        ResourceInProduct resourceInProduct = new ResourceInProduct();
+
+        for (ResourceInProductRequestDto resourceInProductRequestDto : resourcesInProductRequestDto) {
+            Resource resource = resourceRepository.findById(resourceInProductRequestDto.getId())
+                    .orElseThrow(() -> new ResourceNotFoundException(resourceInProductRequestDto.getId()));
```
VladoKat: This will throw an error if you have multiple users owning the same resource. To get a single result, this should be `findByResourceIdAndOwnerId`

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,260 @@
...
+            }
+
+            if (!resourcesInUsers.contains(resourceInUser)) {
+                throw new ResourceInUserNotFoundException(resourceInUser.getResource().getId(), user.getId());
+            }
+
+            double quantity = resourceInProductRequestDto.getQuantity();
+
+            if (resourceInUser.getQuantity() < quantity) {
+                throw new NegativeResourceQuantityException(resourceInUser.getQuantity());
```
VladoKat: This is not a NegativeResourceQuantity. This would be `InsufficientResourceQuantityException`.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,260 @@
...
+
+        resourcesInProducts.forEach(resourceInProduct -> resourceInProduct.setProduct(product));
+        resourceInProductRepository.saveAll(resourcesInProducts);
+
+        return mapToProductResponseDto(product);
+    }
+
+    public List<ProductResponseDto> getAllProducts() {
+        List<Product> products = productRepository.findAll();
+        return products.stream().map(this::mapToProductResponseDto).toList();
```
VladoKat: This method is long and complicated. Try splitting the logic inside it into meaningfully named methods.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,260 @@
...
+        }
+        return resultResourcesInUser;
+    }
+
+    private User getUser(ProductRequestDto productRequestDto) {
+        return userRepository.findById(productRequestDto.getOwnerId())
+                .orElseThrow(() -> new UserNotFoundException(productRequestDto.getOwnerId()));
+    }
+
+    private List<ResourceInProduct> getResourceInProducts(User user, List<ResourceInProductRequestDto> resourcesInProductRequestDto) {
```
VladoKat: This method is very long and complicated as well. Try simplifying it by extracting it into methods. Some checks and fetches might be obsolete here. I will try to simplify it myself but it'd be good if you can give it a go yourself and then we can compare.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,260 @@
...
+                }
+
+                productRepository.deleteById(id);
+            } else {
+                throw new ProductContainsException(id);
+            }
+        } else {
+            throw new ProductIsSoldException(id);
+        }
+    }
```
VladoKat: This method is long as well. Please, try to refactor.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,260 @@
...
+                Product product = productRepository.findById(productId)
+                        .orElseThrow(() -> new ProductNotFoundException(productId));
+                product.setContent(parentProduct);
+                products.add(product);
+            });
+        }
+        return products;
+    }
+
+    private ProductResponseDto mapToProductResponseDto(Product product) {
```
VladoKat: This should be a job for ProductMapper.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,260 @@
...
+        ResourceResponseDto resourceResponseDto = new ResourceResponseDto();
+        if (res == null) {
+            resourceResponseDto = null;
+        } else {
+            resourceResponseDto.setClazz(res.getResource().getClazz());
+            resourceResponseDto.setQuantityType(res.getResource().getQuantityType());
+            resourceResponseDto.setId(res.getResource().getId());
+        }
+        return resourceResponseDto;
+    }
```
VladoKat: Same, here, this should be extracted in a mapper method outside this service.

VladoKat (changes requested):
I managed to look over most of it. I haven't taken a look at the tests. Will do it after addressing current comments. I am trying to refactor it myself but try to do it on your own and we can get together and finalize the solution.
From a glance, I'd say the tests would need heavy refactoring as well. A lot of the object creations can be extracted in TestUtil classes or be taken from already existing ones. You can use the other tests as reference.

src/main/java/jewellery/inventory/dto/response/ProductResponseDto.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/dto/response/resource/ResourceInProductResponseDto.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/invalid_resource_quantity/NegativeResourceQuantityException.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/not_found/ResourceNotFoundException.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
ignatIgnatov: :)

src/main/java/jewellery/inventory/exception/not_found/ResourcesInProductNotFoundException.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/product/ProductContainsException.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/product/ProductContainsException.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/mapper/ProductMapper.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/model/Product.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/model/ResourceInUser.java
ignatIgnatov: JsonIgnore removed!

src/main/java/jewellery/inventory/model/resource/ResourceInProduct.java
ignatIgnatov: JsonIgnore removed!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/repository/ProductRepository.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -30,7 +33,7 @@ public ResponseEntity<Object> handleValidationExceptions(MethodArgumentNotValidE
     return createErrorResponse(HttpStatus.BAD_REQUEST, fieldErrorsMap);
   }
 
-  @ExceptionHandler({NotFoundException.class, ResourceInUserNotFoundException.class})
+  @ExceptionHandler({NotFoundException.class, ResourceInUserNotFoundException.class, ProductWithoutResourcesException.class})
```
VladoKat: Hm, I think ProductWithoutResourcesException should be a 400 BadRequest one because it's a user error that they couldn't supply the required fields. 
Actually, should we provide special error handling for it? If we just make the resourceContent `@NotEmpty` in the ProductRequestDto I think we can get rid of the whole custom validation.

src/main/java/jewellery/inventory/exception/invalid_resource_quantity/NegativeResourceQuantityException.java
```diff
@@ -2,6 +2,6 @@
 
 public class NegativeResourceQuantityException extends InvalidResourceQuantityException {
   public NegativeResourceQuantityException(double quantity) {
-    super("Resource Quantity you entered is negative: " + quantity);
+    super("Resource Quantity you entered is negative " + quantity);
```
VladoKat: If you add ':' to the message as it was, it wouldn't show up as a change. It's good to keep existing correct code not changed to keep better traceability of the actual changes for a PR.

src/main/java/jewellery/inventory/exception/not_found/ProductNotFoundException.java
```diff
@@ -0,0 +1,13 @@
+package jewellery.inventory.exception.not_found;
+
+import java.util.UUID;
+
+public class ProductNotFoundException extends NotFoundException{
+
+    public ProductNotFoundException(UUID id) {
+        super("The product with id " + id + " is not found");
+    }
+    public ProductNotFoundException(String name) {
```
VladoKat: The name one is not used anywhere from what I saw.

src/main/java/jewellery/inventory/exception/not_found/UserNotFoundException.java
```diff
@@ -6,4 +6,8 @@ public class UserNotFoundException extends NotFoundException {
   public UserNotFoundException(UUID id) {
     super("User with id " + id + " was not found");
   }
+
+  public UserNotFoundException(String name) {
+    super("User with name " + name + " was not found");
+  }
```
VladoKat: We don't have this exception thrown anywhere I believe.

src/main/java/jewellery/inventory/exception/product/ProductWithoutResourcesException.java
```diff
@@ -0,0 +1,7 @@
+package jewellery.inventory.exception.product;
+
+public class ProductWithoutResourcesException extends RuntimeException {
+    public ProductWithoutResourcesException() {
+        super("The product cannot be created without resources");
```
VladoKat: If we use `@NotEmpty` for the resourcesContent in the ProductRequestDto, I think this will be handled automatically.

src/main/java/jewellery/inventory/mapper/ProductMapper.java
```diff
@@ -0,0 +1,70 @@
...
+        setContentProductToResponse(product, productResponseDto);
+        setResourcesToResponse(product, productResponseDto);
+        setProductsToResponse(product, productResponseDto);
+
+        return productResponseDto;
+    }
+
+    private void setProductsToResponse(Product product, ProductResponseDto response) {
+        if (product.getProductsContent() == null) {
+            response.setProductsContent(null);
```
VladoKat: The if branch is meaningless. It would be null whether or not you set it to null.

src/main/java/jewellery/inventory/mapper/ProductMapper.java
```diff
@@ -0,0 +1,70 @@
...
+            response.setProductsContent(product.getProductsContent()
+                    .stream().map(this::mapToProductResponseDto)
+                    .toList());
+        }
+    }
+
+    private void setResourcesToResponse(Product product, ProductResponseDto response) {
+        if (product.getResourcesContent() == null) {
+            throw new ProductWithoutResourcesException();
+        } else {
```
VladoKat: This could be handled on Controller level body validation with `@NotEmpty` resourcesContent

src/main/java/jewellery/inventory/mapper/ProductMapper.java
```diff
@@ -0,0 +1,70 @@
...
+            response.setResourcesContent(product.getResourcesContent()
+                    .stream().map(res -> {
+                        ResourceResponseDto resourceResponseDto = resourceMapper.toResourceResponseDto(res);
+
+                        ResourceQuantityResponseDto resourceQuantityResponseDto = new ResourceQuantityResponseDto();
+                        resourceQuantityResponseDto.setResource(resourceResponseDto);
+                        resourceQuantityResponseDto.setQuantity(res.getQuantity());
```
VladoKat: You could extract this into a separate method `mapToResourceQuantityResponse`

src/main/java/jewellery/inventory/mapper/ProductMapper.java
```diff
@@ -0,0 +1,70 @@
...
+                    }).toList());
+        }
+    }
+
+    private void setContentProductToResponse(Product product, ProductResponseDto response) {
+        if (product.getContentOf() == null) {
+            response.setContentOf(null);
+        } else {
+            response.setContentOf(product.getContentOf().getId());
+        }
```
VladoKat: would be the same if you did it like this:
```
private void setContentProductToResponse(Product product, ProductResponseDto response) {
    if (product.getContentOf() != null) {
        response.setContentOf(product.getContentOf().getId());
    }
}
```

src/main/java/jewellery/inventory/mapper/ResourceMapper.java
```diff
@@ -43,4 +43,13 @@ public Resource toResourceEntity(ResourceRequestDto resourceRequestDto) {
...
   }
+
+  public ResourceResponseDto toResourceResponseDto(ResourceInProduct resource) {
+    ResourceResponseDto resourceResponseDto = new ResourceResponseDto();
+    resourceResponseDto.setClazz(resource.getResource().getClazz());
+    resourceResponseDto.setQuantityType(resource.getResource().getQuantityType());
+    resourceResponseDto.setId(resource.getResource().getId());
+
+    return resourceResponseDto;
+  }
```
VladoKat: Why do you need this method? You should be able to map Resource to ResourceResponseDto. If you want to do that for a ResourceInProduct, you could just do `toResourceResponse(resourceInProduct.getResource())`

src/main/java/jewellery/inventory/model/resource/ResourceInProduct.java
VladoKat: Is it? It's still showing it.

src/main/java/jewellery/inventory/repository/ResourceInUserRepository.java
```diff
@@ -12,4 +12,8 @@ public interface ResourceInUserRepository extends JpaRepository<ResourceInUser,
   @Query(
       "SELECT COALESCE(SUM(riu.quantity), 0.0) FROM ResourceInUser riu WHERE riu.resource.id = :resourceId")
   Double sumQuantityByResource(@Param("resourceId") UUID resourceId);
+
+  ResourceInUser findByResourceId(UUID resourceId);
```
VladoKat: This should be removed as it might mislead someone and we shouldn't be using it.

src/main/java/jewellery/inventory/repository/ResourceRepository.java
VladoKat, line 1: This is an obsolete change. I would revert it to keep the PR clean.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,230 @@
...
+        List<ResourceInProduct> resourcesInProduct = getResourcesFromRequest(user, productRequestDto.getResourcesContent());
+
+        Product product = setFieldsToNewProduct(productRequestDto, user, resourcesInProduct);
+        productRepository.save(product);
+
+        setContentProduct(productRequestDto, product);
+        setProductToResourcesInProduct(resourcesInProduct, product);
+
+        return productMapper.mapToProductResponseDto(product);
+    }
```
VladoKat: I came up with a similarly structured code. Check it out.
```
  public ProductResponseDto createProduct(ProductRequestDto productRequestDto) {

    User owner = getUser(productRequestDto.getOwnerId());

    Product product = createProductWithoutResourcesAndProducts(productRequestDto, owner);

    addProductContentToProduct(productRequestDto, product);

    addResourcesToProduct(productRequestDto, owner, product);

    return mapToProductResponseDto(product);
  }
```
The main difference is I add the resources to the product after I've created the product. That way I can set the product directly to the ResourceInProduct objects and we don't need to pass the resourcesInProduct in 2 places.

VladoKat: Another think I just thought of is the user should be able to add products that they themselves own. I don't think we are checking for that anywhere.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,230 @@
...
+
+    private void disassembleProductContent(Product product) {
+        if (product.getProductsContent() != null) {
+            product.getProductsContent().forEach(content -> {
+                content.setContentOf(null);
+                productRepository.save(content);
+            });
+
+            product.setProductsContent(new ArrayList<>());
+            productRepository.save(product);
```
VladoKat: Have you tried to just do this? 
```
product.setProductsContent(new ArrayList<>());
productRepository.save(product);
```
I am not sure how smart JPA is but I think with the configuration we have given our Product entity if you just save the product without content, it should update the contentOf fields of all related products? (I haven't tried it. Just wondering)

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,230 @@
...
+            ResourceInUser resourceInUser = moveResourcesToUserWhenDisassembleProduct(owner, resourceInProduct, resource);
+
+            resultResourcesInUser.add(resourceInUser);
+
+            if (resourceInProduct.getProduct() == null) {
+                resourceInProductRepository.deleteById(resourceInProduct.getId());
+            }
+        });
+
+        resourceInUserRepository.saveAll(resultResourcesInUser);
```
VladoKat: Instead of this, can we try to utilise ResourceInUserService::addResourceToUser? It should be nice to reuse functionality instead of writing it anew. We'd need to map the ResourceInProduct to ResourceInUserRequest, but that should be easy enough.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,230 @@
...
+        if (resourceInUser == null) {
+            resourceInUser = new ResourceInUser();
+            resourceInUser.setResource(resource);
+            resourceInUser.setQuantity(resourceInProduct.getQuantity());
+            resourceInUser.setOwner(owner);
+        } else {
+            resourceInUser.setQuantity(resourceInUser.getQuantity() + resourceInProduct.getQuantity());
+        }
+        return resourceInUser;
+    }
```
VladoKat: If we utilise ResourceInUserService::addResourceToUser, then this method would be obsolete.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,230 @@
...
+
+    private User getUserFromRequest(ProductRequestDto productRequestDto) {
+        return userRepository.findById(productRequestDto.getOwnerId())
+                .orElseThrow(() -> new UserNotFoundException(productRequestDto.getOwnerId()));
+    }
+
+    private List<ResourceInProduct> getResourcesFromRequest(User user, List<ResourceQuantityRequestDto> resourcesInProductRequestDto) {
+        if (resourcesInProductRequestDto == null) {
+            throw new ProductWithoutResourcesException();
+        }
```
VladoKat: This would be removed if we used `@NotEmpty` on resourcesContent in the ProductRequestDto.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,230 @@
...
+        List<ResourceInUser> resourcesInUsers = user.getResourcesOwned();
+        if (resourcesInUsers == null) {
+            throw new ResourceInUserNotFoundException(user.getId());
+        }
```
VladoKat: I am not sure we should be doing this check. Also, would this ever be null? I imagine it should be an empty list if there are no Resources in the user.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,230 @@
...
+
+        resourceInUserRepository.saveAll(resultResourcesInUser);
+    }
+
+    private User getUserFromRequest(ProductRequestDto productRequestDto) {
+        return userRepository.findById(productRequestDto.getOwnerId())
+                .orElseThrow(() -> new UserNotFoundException(productRequestDto.getOwnerId()));
+    }
+
+    private List<ResourceInProduct> getResourcesFromRequest(User user, List<ResourceQuantityRequestDto> resourcesInProductRequestDto) {
```
VladoKat: This method is doing a lot more than just getting the resources from a request. It is also removing quantities from user and adding them to products. This functionality shouldn't be part of this method.
Let's discuss how we can improve this together.

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
```diff
@@ -0,0 +1,66 @@
...
+        resourceInUser.setOwner(user);
+        resourceInUser.setResource(pearl);
+        resourceInUser.setQuantity(20);
+        return resourceInUser;
+    }
+
+    public static ProductRequestDto createProductRequestDto(User user, ResourceQuantityRequestDto resourceQuantityRequestDto) {
+        ProductRequestDto productRequestDto = new ProductRequestDto();
+        productRequestDto.setOwnerId(user.getId());
+        productRequestDto.setName("TestProductDto");
```
VladoKat: I believe the productRequestDto and the Product entity should hold the same properties for most of the test cases. Otherwise you wouldn't be able to compare them properly afterwards. Consider extracting common values like name, description, salePrice to constants.

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,232 @@
...
+        return "http://localhost:" + port + "/resources";
+    }
+
+    private String getBaseUserUrl() {
+        return "http://localhost:" + port + "/users";
+    }
+
+    private String getBaseProductUrl() {
+        return "http://localhost:" + port + "/products";
+    }
```
VladoKat: These could make use of getBaseUrl method.

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,232 @@
...
+    void createProductSuccessfully() {
+
+        ResponseEntity<ProductResponseDto> response =
+                this.testRestTemplate.postForEntity(getBaseProductUrl(), productRequestDto, ProductResponseDto.class);
+        ProductResponseDto productResponseDto = response.getBody();
+
+        assertEquals(HttpStatus.CREATED, response.getStatusCode());
+
+        assertEquals(productRequestDto.getName(), Objects.requireNonNull(productResponseDto).getName());
+        assertEquals(productRequestDto.getResourcesContent().get(0).getId(), productResponseDto.getResourcesContent().get(0).getResource().getId());
```
VladoKat: keep in mind we should strive for the following structure in our tests:
```
void test(){
<preparation>

<execution>

<assertions>
}
```

In this specific there is no preparation part but we can't make a clear distinction where the execution and the assertions are.

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+    @Test
+    void getProductSuccessfully() {
+
+        productResponseDto = getProductResponseDto(productRequestDto);
+
+        ResponseEntity<ProductResponseDto> response =
+                this.testRestTemplate.getForEntity(getProductUrl(Objects.requireNonNull(productResponseDto).getId()), ProductResponseDto.class);
+
+        assertEquals(HttpStatus.OK, response.getStatusCode());
```
VladoKat: We should add more assertions for the result of a test. We should somehow compare the object that's coming back from the response with what we expect.

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,232 @@
...
+    @Test
+    void getAllProductsSuccessfully() {
+        productResponseDto = getProductResponseDto(productRequestDto);
+
+        ResponseEntity<List<ProductResponseDto>> response =
+                this.testRestTemplate.exchange(
+                        getBaseProductUrl(), HttpMethod.GET, null, new ParameterizedTypeReference<>() {
+                        });
+
+        assertEquals(HttpStatus.OK, response.getStatusCode());
```
VladoKat: Same here, we should validate that there is a product created in the db with the properties we have put in the request.

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,232 @@
...
+        ProductRequestDto productRequestDto = new ProductRequestDto();
+        productRequestDto.setName("TestProduct");
+        productRequestDto.setAuthors(List.of("TestAuthors"));
+        productRequestDto.setDescription("Test");
+        productRequestDto.setOwnerId(user.getId());
+        productRequestDto.setSalePrice(50);
+        productRequestDto.setResourcesContent(listOfResourcesInProduct);
+
+        return productRequestDto;
+    }
```
VladoKat: Should we move this to the helper as well?

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,232 @@
...
+        return resourceInUserRequestDto;
+    }
+
+    @Nullable
+    private Gemstone getGemstone() {
+        ResourceRequestDto resourceRequest = ResourceTestHelper.getGemstoneRequestDto();
+        ResponseEntity<Gemstone> createResource =
+                this.testRestTemplate.postForEntity(getBaseResourceUrl(), resourceRequest, Gemstone.class);
+
+        return createResource.getBody();
```
VladoKat: This is a bad name for this method. We are not just getting/creating it as a pojo. This is creating a gemstone in the DB. Furthermore, the response from creating a resource is not a Resource, it's ResourceResponseDto. This would fail if the Dto and the entity have different fields.

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,232 @@
...
+        });
+        return listOfResourcesInProduct;
+    }
+
+    @Nullable
+    private ResourcesInUserResponseDto getResourcesInUserResponseDto(ResourceInUserRequestDto resourceInUserRequestDto) {
+        ResponseEntity<ResourcesInUserResponseDto> createResourceInUser =
+                this.testRestTemplate.postForEntity(getBaseResourceAvailabilityUrl(), resourceInUserRequestDto, ResourcesInUserResponseDto.class);
+
+        System.out.println(createResourceInUser.getBody().getOwner());
```
VladoKat: Please, make sure we are not keeping any System.out in the code that gets merged.

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+        return createResource.getBody();
+    }
+
+    @Nullable
+    private User getUser() {
+        UserRequestDto userRequest = UserTestHelper.createTestUserRequest();
+        ResponseEntity<User> createUser = this.testRestTemplate.postForEntity(getBaseUserUrl(), userRequest, User.class);
+
+        return createUser.getBody();
```
VladoKat: Same thing goes here.

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
VladoKat, line 1: Have you changed anything here or is it some formatting change? Please revert it if it's the latter.

VladoKat (changes requested):
Take a look and let's discuss the refactor of the ProductService.

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
ignatIgnatov: I removed this exception everywhere and made ResourceContent @NotNull

src/main/java/jewellery/inventory/exception/not_found/ProductNotFoundException.java
ignatIgnatov: Yes, I had made a change to the code and forgot to remove this exception. I have removed it now!

src/main/java/jewellery/inventory/exception/not_found/UserNotFoundException.java
ignatIgnatov: Removed!

src/main/java/jewellery/inventory/exception/product/ProductWithoutResourcesException.java
ignatIgnatov: Yes, I did it

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: createGemstoneInDatabase?

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: createUserInDatabase?

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
ignatIgnatov: I haven't made any changes to UserCrudIntegrationTest.

src/main/java/jewellery/inventory/exception/product/ProductWithoutResourcesException.java
VladoKat: `NotNull` is not enough in this case because resourcesContent can also be an empty array. This is not good for us as well. That's why I suggested `NotEmpty`. This checks for both null and empty array

src/main/java/jewellery/inventory/dto/request/ProductRequestDto.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.List;
+import java.util.UUID;
+
+@Getter
+@Setter
+@RequiredArgsConstructor
+@AllArgsConstructor
+public class ProductRequestDto {
+
+  @NotNull
```
VladoKat: This should be `@NotEmpty` instead of `@NotNull`. The difference is the first one will fail also if the list is initialized as a list but has no elements.

src/main/java/jewellery/inventory/dto/request/ProductRequestDto.java
```diff
@@ -0,0 +1,25 @@
...
+  private List<UUID> authors;
+  private UUID ownerId;
+  @NotNull List<ResourceQuantityRequestDto> resourcesContent;
+  private List<UUID> productsContent;
+  private String description;
+  private double salePrice;
+  private String catalogNumber;
+  private String productionNumber;
```
VladoKat: I'm thinking, given this is a requestDto, maybe we should introduce at least `@NotNull `validation for the required fields - ownerId, salePrice, catalogNumber, productNumber

src/main/java/jewellery/inventory/mapper/ProductMapper.java
```diff
@@ -0,0 +1,83 @@
...
+                })
+            .toList());
+  }
+
+  private void setContentProductToResponse(Product product, ProductResponseDto response) {
+    if (product.getContentOf() == null) {
+      response.setContentOf(null);
+    } else {
+      response.setContentOf(product.getContentOf().getId());
+    }
```
VladoKat: This could become:
```
if(product.getContentOf() != null){
 response.setContentOf(product.getContentOf().getId());
}
```

src/main/java/jewellery/inventory/mapper/ProductMapper.java
```diff
@@ -0,0 +1,83 @@
...
+    return response;
+  }
+
+  private void setProductsToResponse(Product product, ProductResponseDto response) {
+    if (product.getProductsContent() == null) {
+      response.setProductsContent(null);
+    } else {
+      response.setProductsContent(
+          product.getProductsContent().stream().map(this::mapToProductResponseDto).toList());
+    }
```
VladoKat: The first branch is obsolete. You could just go for:
```
if(product.getProductsContent() != null){
   response.setProductsContent(
          product.getProductsContent().stream().map(this::mapToProductResponseDto).toList());
}
```

src/main/java/jewellery/inventory/mapper/ProductMapper.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  private void setResourcesToResponse(Product product, ProductResponseDto response) {
+
+    response.setResourcesContent(
+        product.getResourcesContent().stream()
+            .map(
+                res -> {
+                  ResourceResponseDto resourceResponseDto =
+                      resourceMapper.toResourceResponseDto(res);
```
VladoKat: Rename res to be resourceAndQuantity and then you can just use resourceMapper.toResourceResponseDto(resourceAndQuantity.getResource())

src/main/java/jewellery/inventory/mapper/ResourceMapper.java
VladoKat: This is an obsolete method.

src/main/java/jewellery/inventory/model/User.java
```diff
@@ -14,6 +14,9 @@
 public class User {
   @Id @GeneratedValue private UUID id;
 
+  @ManyToMany(mappedBy = "authors")
+  private List<Product> createdProducts;
```
VladoKat: let's name this authoredProducts to have a clear distinction between author and owner. Created products can be mistaken for productsOwned

src/main/java/jewellery/inventory/model/Product.java
```diff
@@ -1,29 +1,44 @@
...
 import lombok.Getter;
+import lombok.RequiredArgsConstructor;
 import lombok.Setter;
 
 @Entity
 @Getter
 @Setter
-public class Product {
+@RequiredArgsConstructor
+public class Product implements Serializable {
```
VladoKat: Do we need to explicitly mark it as Serializable?

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,223 @@
...
+        productRepository.findById(id).orElseThrow(() -> new ProductNotFoundException(id));
+
+    throwExceptionIfProductIsSold(id, product);
+    throwExceptionIfProductIsPartOfAnotherProduct(id, product);
+
+    moveResourceInProductToResourceInUser(product);
+    disassembleProductContent(product);
+
+    productRepository.deleteById(id);
+  }
```
VladoKat: Can we move all the public methods on top to keep some sort of structure.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,223 @@
...
+  private List<User> getAuthors(ProductRequestDto productRequestDto) {
+    List<UUID> authorsIds = productRequestDto.getAuthors();
+    List<User> authors = new ArrayList<>();
+    authorsIds.forEach(
+        id -> {
+          User author =
+              userRepository.findById(id).orElseThrow(() -> new UserNotFoundException(id));
+          authors.add(author);
+        });
+    return authors;
```
VladoKat: The solution is good as it is 👍 . Just recommending a "one-liner" in functional style.
```suggestion
    return productRequestDto.getAuthors().stream()
        .map(
            authorId ->
                userRepository
                    .findById(authorId)
                    .orElseThrow(() -> new UserNotFoundException(authorId)))
        .toList();
```

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  private void addProductsContentToProduct(ProductRequestDto productRequestDto, Product product) {
+    if (productRequestDto.getProductsContent() == null) {
+      product.setProductsContent(null);
+    } else {
```
VladoKat: The if part is obsolete. You can inverse the condition and just do logic in the else branch.

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  public static ProductRequestDto createProductRequestDto(
+      User user, ResourceQuantityRequestDto resourceQuantityRequestDto) {
+    ProductRequestDto productRequestDto = new ProductRequestDto();
+    productRequestDto.setAuthors(List.of(user.getId()));
+    productRequestDto.setOwnerId(user.getId());
+    productRequestDto.setDescription("This is test product");
+    productRequestDto.setResourcesContent(List.of(resourceQuantityRequestDto));
+    productRequestDto.setSalePrice(10000);
```
VladoKat: We should also provide values for the new fields as they are going to be mandatory.

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
```diff
@@ -0,0 +1,100 @@
...
+
+  @NotNull
+  public static ProductRequestDto getProductRequestDto(
+      ResourcesInUserResponseDto resourcesInUser, User user) {
+
+    List<ResourceQuantityRequestDto> listOfResourcesInProduct =
+        getResourceInProductRequestDtos(resourcesInUser);
+
+    ProductRequestDto productRequestDto = new ProductRequestDto();
+    productRequestDto.setDescription("Test");
```
VladoKat, line 51: I imagine we are going to compare the description from the request to the one we are saving in the DB. This means we should use the same description in the request and in the entity.

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
```diff
@@ -0,0 +1,100 @@
...
+    productRequestDto.setOwnerId(user.getId());
+    productRequestDto.setAuthors(List.of(user.getId()));
+    productRequestDto.setSalePrice(50);
```
VladoKat, line 54: Same for the salePrice?

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
```diff
@@ -0,0 +1,100 @@
...
+    productRequestDto.setAuthors(List.of(user.getId()));
+    productRequestDto.setOwnerId(user.getId());
+    productRequestDto.setDescription("This is test product");
+    productRequestDto.setResourcesContent(List.of(resourceQuantityRequestDto));
+    productRequestDto.setSalePrice(10000);
+    return productRequestDto;
+  }
+
+  @NotNull
+  public static ProductRequestDto getProductRequestDto(
```
VladoKat, line 42: I think you should use the same naming convention for creating ProductRequestDto. In this case I would go for renaming the first method to be getProductRequestDto as well. This way you would overload the name and when we want to create a ProductRequestDto and we type getProductRequestDto it will show us both methods and depending on what we have in our scenario we would choose which method to use. Keep in mind, we expect them to behave more or less the same - the fields that are not provided (description, salePrice) should have the same values.

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
VladoKat: ^

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,199 @@
...
+    ProductRequestDto productRequestDto;
+    ProductResponseDto productResponseDto;
+
+    @BeforeEach
+    void setUp() {
+        userRepository.deleteAll();
+        productRepository.deleteAll();
+        resourceRepository.deleteAll();
+        resourceInUserRepository.deleteAll();
+        resourceInProductRepository.deleteAll();
```
VladoKat: extract in cleanupDb()?

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,199 @@
...
+    ResourceRepository resourceRepository;
+    @Autowired
+    ResourceInUserRepository resourceInUserRepository;
+    @Autowired
+    ResourceInProductRepository resourceInProductRepository;
+
+    User user;
+    Gemstone gemstone;
+    ResourceInUserRequestDto resourceInUserRequestDto;
+    ResourcesInUserResponseDto resourcesInUser;
```
VladoKat: maybe rename to resourcesInUserResponseDto to keep the naming convention clear. I thought this is an entity just by looking at the name.

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,199 @@
...
+        productResponseDto = getProductResponseDto(productRequestDto);
+
+        ResponseEntity<HttpStatus> response =
+                this.testRestTemplate.exchange(
+                        getProductUrl(Objects.requireNonNull(productResponseDto).getId()),
+                        HttpMethod.DELETE,
+                        null,
+                        HttpStatus.class);
+
+        assertEquals(HttpStatus.NO_CONTENT, response.getStatusCode());
```
VladoKat: We can make sure that there is no product in DB here.

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  @Test
+  void testDeleteProductSuccessfully() {
+
+    when(productRepository.findById(testProduct.getId())).thenReturn(Optional.of(testProduct));
+
+    productService.deleteProduct(testProduct.getId());
+
+    assertEquals(0, productRepository.count());
```
VladoKat: productRepository is mocked. I believe this is always going to be 0. This test is obsolete.

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  @Test
+  void testDeleteProductContainsProductBreakDown() {
+    testProduct.setProductsContent(List.of(new Product(), new Product(), new Product()));
+    when(productRepository.findById(testProduct.getId())).thenReturn(Optional.of(testProduct));
+
+    productService.deleteProduct(testProduct.getId());
+
+    assertEquals(0, testProduct.getProductsContent().size());
```
VladoKat: You could also set contentOf of these products to be the testProduct. Then you can assert that after deletion they no longer have the contentOf fields set. This would be a more complete productBreakdown test.

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  @Test
+  void testDeleteProductGetProductShouldThrowExceptionWhenProductNotExist() {
+    UUID fakeId = UUID.randomUUID();
+    assertThrows(ProductNotFoundException.class, () -> productService.deleteProduct(fakeId));
+  }
+
+  @Test
+  void createProductSuccessfully() {
```
VladoKat, line 65: I am wondering whether we should delete this test as we are testing it in the integration tests.

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+    when(resourceInUserRepository.findByResourceIdAndOwnerId(pearl.getId(), user.getId()))
+        .thenReturn(Optional.of(resourceInUser));
+    resourceInUser.setQuantity(5);
+    user.setResourcesOwned(List.of(resourceInUser));
+    resourceQuantityRequestDto.setQuantity(5);
+
+    productService.createProduct(productRequestDto);
+
+    assertEquals(0, resourceInUserRepository.count());
+  }
```
VladoKat: This is not needed considering we are utilizing the resourceInUserService's functionality and it would be mocked here. Furthermore, resourceInUserRepository i smocked so unless we stub it, it's always going to return 0.

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+        () -> productService.createProduct(productRequestDto));
+  }
+
+  @Test
+  void testCreateProductShouldThrowWhenResourceNotExistInResourceInUserRepository() {
+    when(userRepository.findById(user.getId())).thenReturn(Optional.of(user));
+    assertThrows(
+        ResourceInUserNotFoundException.class,
+        () -> productService.createProduct(productRequestDto));
+  }
```
VladoKat: This should also be deleted? Is it different than the test above?

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+    when(productRepository.findById(testProduct.getId())).thenReturn(Optional.of(testProduct));
+
+    productRequestDto.setProductsContent(List.of(testProduct.getId()));
+
+    ProductResponseDto response = new ProductResponseDto();
+    when(productMapper.mapToProductResponseDto(any())).thenReturn(response);
+
+    ProductResponseDto actual = productService.createProduct(productRequestDto);
+
+    assertEquals(response, actual);
```
VladoKat: The assertion here is the same as in the createProductSuccessfully. We should make extra assertions for the product.getContentOf to make sure it is set

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  @Test
+  void testCreateProductShouldThrowWhenProductNotFound() {
```
VladoKat: rename to `testCreateProductShouldThrowWhenProductNotFound`. Also this seems to be the same as `testCreateProductShouldThrowExceptionWhenProductToContainNotFound`. Remove one of them.

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+    when(userRepository.findById(user.getId())).thenReturn(Optional.of(user));
+    user.setResourcesOwned(List.of(resourceInUser));
+    productRequestDto.setProductsContent(List.of(UUID.randomUUID()));
+
+    assertThrows(
+        ProductNotFoundException.class, () -> productService.createProduct(productRequestDto));
+  }
+
+  @Test
+  void testCreateProductShouldThrowExceptionWhenResourceNotFound() {
```
VladoKat, line 129: What's the difference between this and `testCreateProductShouldThrowExceptionWhenResourceInUserNotFound`. Please, remove this one if they are the same

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,264 @@
...
+  void testGetAllProducts() {
+
+    List<Product> products = Arrays.asList(testProduct, new Product(), new Product());
+
+    when(productRepository.findAll()).thenReturn(products);
+
+    List<ProductResponseDto> responses = productService.getAllProducts();
+
+    assertEquals(products.size(), responses.size());
+  }
```
VladoKat: These may be obsolete if we are testing them properly in the intergation/feature tests.

VladoKat (changes requested):
Left mainly minor comments. Please, check the tests you have made for any repeating code and try to extract the common in methods (e.g. same stubs in multiple tests). Try to improve the checks we are doing in the feature tests.

src/main/java/jewellery/inventory/model/resource/ResourceInProduct.java
ignatIgnatov: Now it is! :)

src/main/java/jewellery/inventory/exception/invalid_resource_quantity/NegativeResourceQuantityException.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/product/ProductWithoutResourcesException.java
ignatIgnatov: I did the following: 
@NotEmpty
  private List<UUID> authors;
  @NotNull
  private UUID ownerId;
  @NotEmpty
  List<ResourceQuantityRequestDto> resourcesContent;

src/main/java/jewellery/inventory/mapper/ResourceMapper.java
ignatIgnatov: I have removed this method. Now I use toResourceResponse().

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: It's done!

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/mapper/ProductMapper.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/mapper/ProductMapper.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/model/User.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/model/Product.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ProductService.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
ignatIgnatov: We already discussed this, I will rename testProduct to testContentProduct

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
ignatIgnatov: We already discussed this

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
ignatIgnatov: We already discussed this

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
ignatIgnatov: Removed!

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
ignatIgnatov: Removed!

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
ignatIgnatov: Done! But I think that in this test something smells... Maybe I want to remove it...

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
ignatIgnatov: I did more assertions in integration test

src/main/java/jewellery/inventory/dto/request/ProductRequestDto.java
```diff
@@ -0,0 +1,28 @@
...
+  private List<UUID> authors;
+  @NotNull
+  private UUID ownerId;
+  @NotEmpty
+  List<ResourceQuantityRequestDto> resourcesContent;
+  private List<UUID> productsContent;
+  private String description;
+  private double salePrice;
+  private String catalogNumber;
+  private String productionNumber;
```
VladoKat: I haven't mentioned it before but can we put `@NotNull` on these as well? (only description is not mandatory)

src/main/java/jewellery/inventory/mapper/ResourceMapper.java
```diff
@@ -42,5 +42,6 @@ public Resource toResourceEntity(ResourceRequestDto resourceRequestDto) {
       return linkingPartMapper.toResourceEntity(linkingPartDTO);
     }
     throw new MappingException(resourceRequestDto);
+
```
VladoKat: obsolete change

src/main/java/jewellery/inventory/model/resource/ResourceInProduct.java
```diff
@@ -15,9 +15,11 @@
...
 
-  @ManyToOne private Resource resource;
+  @ManyToOne
+  private Resource resource;
 
   private double quantity;
 
-  @ManyToOne private Product product;
+  @ManyToOne
+  private Product product;
```
VladoKat: These are obsolete changes as well.

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
```diff
@@ -0,0 +1,106 @@
...
+import jewellery.inventory.model.resource.ResourceInProduct;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+public class ProductTestHelper {
+
+  public static ResourceInUser createResourceInUser(User user, Resource pearl) {
```
VladoKat: I'd say if we are going for get<dto/model> then we should stick to that for all of the dto/model object creation.

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -0,0 +1,255 @@
...
+    @Mock
+    UserRepository userRepository;
+    @Mock
+    ResourceRepository resourceRepository;
+    @Mock
+    ResourceInUserRepository resourceInUserRepository;
+    @Mock
+    ResourceInProductRepository resourceInProductRepository;
+    @Mock
+    ResourceInUserService resourceInUserService;
```
VladoKat: Those should also be private

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -0,0 +1,222 @@
...
+    ResourceInUserRepository resourceInUserRepository;
+    @Autowired
+    ResourceInProductRepository resourceInProductRepository;
+
+    User user;
+    Gemstone gemstone;
+    ResourceInUserRequestDto resourceInUserRequestDto;
+    ResourcesInUserResponseDto resourcesInUserResponseDto;
+    ProductRequestDto productRequestDto;
+    ProductResponseDto productResponseDto;
```
VladoKat: Those should be private

VladoKat (changes requested):
Last minor changes.

src/main/java/jewellery/inventory/dto/request/ProductRequestDto.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/model/resource/ResourceInProduct.java
ignatIgnatov: yes

src/main/java/jewellery/inventory/mapper/ResourceMapper.java
ignatIgnatov: yes

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
ignatIgnatov: Done!

VladoKat (approved):
LGTM!

PR #41, opened 2023-10-23
Feature 31 resource model restructure

ignatIgnatov:


VladoKat (commented):
PR looks good but we should merge it when the frontend is ready for these changes as well -> https://trello.com/c/HCjgN9OF/33-fe-resource-columns-change

VladoKat (changes requested):
Can you include also the script for migrating the database according to the new version of the resource objects

VladoKat (approved):
https://github.com/JewelleryManagement/jewellery-management-ui/pull/36 merged. We can move this forward as well

PR #43, opened 2023-10-25
Feature 36 init flyway

ignatIgnatov:


src/main/resources/db/migration/V1_1__insert_user_admin.sql
VladoKat, line 1: Should we update the README to remove the data.sql file and mention that we are creating the root user through the code

VladoKat (changes requested):

src/main/resources/db/migration/V1_1__insert_user_admin.sql
ignatIgnatov: Done!

VladoKat (approved):

PR #47, opened 2023-11-06
Feature 34 product sale

berki0:


src/main/java/jewellery/inventory/dto/request/ProductPriceDiscountRequestDto.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.UUID;
+
+@Getter
+@Setter
+@RequiredArgsConstructor
+@AllArgsConstructor
+public class ProductPriceDiscountRequestDto {
+  @NotNull private UUID productId;
+  @NotNull private double salePrice;
+  @NotNull private double discount;
```
VladoKat: If we use double it can never be null. I don't know how it would behave if we didn't pass it in the body in this case. I'd imagine it would have value 0. Check wrapper vs primitive comparison -> https://www.geeksforgeeks.org/java-i-o-operation-wrapper-class-vs-primitive-class-variables/

src/main/java/jewellery/inventory/dto/response/ProductResponseDto.java
```diff
@@ -23,4 +23,5 @@ public class ProductResponseDto {
   private boolean isSold;
```
VladoKat: we should include the sale information here. We can do that by replacing the isSold wit partOfSale with the id of the sale entity

src/main/java/jewellery/inventory/exception/product/ProductOwnerNotSeller.java
```diff
@@ -0,0 +1,11 @@
+package jewellery.inventory.exception.product;
+
+import java.util.UUID;
+
+public class ProductOwnerNotSeller extends RuntimeException {
```
VladoKat: I'd rename it to SellerNotOwner but either way the exception name should end with Exception. (SellerNotOwnerException)

src/main/java/jewellery/inventory/mapper/SaleMapper.java
```diff
@@ -0,0 +1,128 @@
...
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class SaleMapper {
+  private final UserMapper userMapper;
+  private final ProductMapper productMapper;
+  private final UserRepository userRepository;
+  private final ProductRepository productRepository;
```
VladoKat: Mappers shouldn't use repositories. Getting values from the database shouldn't be the concern of a mapper. Mapper should only take passed arguments and turn them into another object. It should not be coupled with repository logic. All the fields needed for mapping of one entity to another should be passed to the mapper method.

src/main/java/jewellery/inventory/mapper/SaleMapper.java
```diff
@@ -0,0 +1,128 @@
...
+            .toList();
+    return setProductPriceAndDiscount(saleRequestDto, productList);
+  }
+
+  private double getDiscountInPercentage(List<ProductPriceDiscountRequestDto> discountList) {
+    double discountPercentage = 0;
+    for (ProductPriceDiscountRequestDto productPriceDiscountRequestDto : discountList) {
+      discountPercentage += productPriceDiscountRequestDto.getDiscount();
+    }
+    return discountPercentage / discountList.size();
```
VladoKat: This calculation is not correct. Example: you have 2 items worth 1000 and 10. You make 10% discount on the 1000 and 90% on the 10. That means the total amount is 1010, the discounted amount is 901 = (90% * 1000) + (10% * 10) . Let's name the discounted amount in percentage as x. Then ((x/100) * totalAmount = discountedAmount) =>  (x / 100) * 1010 = 901 => x = (901 / 1010) * 100 => x ~= 89.2% => the total discount is 10.8%. By your calculation it would be (10 + 90) / 2 = 50%. The calculation is wrong because the discounts are applied to different prices so they cannot be calculated by averaging.

src/main/java/jewellery/inventory/mapper/SaleMapper.java
```diff
@@ -0,0 +1,128 @@
...
+
+  private double getTotalDiscountFromEntity(List<Product> products) {
+    double totalDiscountAmount = 0;
+    double totalPrice = 0;
+
+    for (Product product : products) {
+      double discountAmount = product.getSalePrice() * (product.getDiscount() / 100);
+      totalDiscountAmount += discountAmount;
+      totalPrice += product.getSalePrice();
+    }
```
VladoKat: We have code repetition from the previous method. Perhaps these could be extracted?

src/main/java/jewellery/inventory/mapper/SaleMapper.java
```diff
@@ -0,0 +1,128 @@
...
+    if (totalPrice != 0) {
+      return (totalDiscountAmount / totalPrice) * 100;
```
VladoKat, line 70: This would be the percentage value of the discountedAmount from the totalPrice. To get the discount percentage you should do
```suggestion
      return 100 - (totalDiscountAmount / totalPrice) * 100;
```

src/main/java/jewellery/inventory/service/SaleService.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public List<SaleResponseDto> getAllSales() {
+    List<Sale> sales = saleRepository.findAll();
+    return sales.stream().map(saleMapper::mapEntityToResponseDto).toList();
+  }
+
+  public SaleResponseDto createSale(SaleRequestDto saleRequestDto) {
+    Sale sale = saleMapper.mapRequestToEntity(saleRequestDto);
+    throwExceptionIfProductOwnerNotSeller(sale.getProducts(), saleRequestDto.getSellerId());
+    throwExceptionProductOwnerEqualsRecipientException(
```
VladoKat: We actually have a business case in which the seller would sell the products to themselves. Example is if in real life a distributor sells their products without disclosing to whom they've sold the products to. Then it'd be easier for us to mark the buyer and seller as the same user.

src/main/resources/db/migration/V3__update_discount_values_in_product.sql
VladoKat, line 1: This should be a part of the same migration file

src/main/resources/db/migration/V4__delete_isSold_field_in_product.sql
VladoKat, line 1: This should also be a part of the same migration

src/main/resources/db/migration/V5__add_partOfSale_in_table_product.sql
VladoKat, line 1: This one as well. We should include all of the changes related to the sales change in a single migration file. It would make sense to have them separate if each one was describing a different state of the backend but in this case we don't have a state of the backend that requires the changes up to, let's say, V4 but not V5.

src/main/java/jewellery/inventory/model/Sale.java
```diff
@@ -0,0 +1,25 @@
...
+public class Sale {
+  @Id @GeneratedValue private UUID id;
+  @ManyToOne private User seller;
+  @ManyToOne private User buyer;
+
+  @OneToMany(mappedBy = "partOfSale")
+  private List<Product> products;
+
+  private Date date;
+  private double discount;
```
VladoKat: I know AC says we should have the discount as part of the Sale but after we added discount to the Product, then the total discount should be calculated via all the products' discounts and doesn't make sense to save it in the sale entity. It should only be calculated in the response we send back but we shouldn't save it in the database.

src/test/java/jewellery/inventory/helper/SaleTestHelper.java
```diff
@@ -0,0 +1,75 @@
...
+    productRequest.setDiscount(discount);
+    return productRequest;
+  }
+  public static List<SaleResponseDto> getSaleResponseList(SaleResponseDto saleResponseDto) {
+    List<SaleResponseDto> saleResponseDtoList = new ArrayList<>();
+    saleResponseDtoList.add(saleResponseDto);
+    return saleResponseDtoList;
+  }
+
+  public static List<Product> getProList(Product product) {
```
VladoKat: This name is a bit weird. Perhaps a typo?

src/test/java/jewellery/inventory/helper/SaleTestHelper.java
```diff
@@ -0,0 +1,75 @@
...
+    List<Product> products = new ArrayList<>();
+    products.add(product);
+    return products;
+  }
+  public static SaleResponseDto getSaleResponseDto(Sale sale) {
+    SaleResponseDto dto = new SaleResponseDto();
+    UserResponseDto userRequestDtoSeller = new UserResponseDto();
+    UserResponseDto userRequestDtoBuyer = new UserResponseDto();
```
VladoKat: The type is UserResponseDto but the variables are userRequestDtoSeller/Buyer. Perhaps should be renamed.

src/test/java/jewellery/inventory/helper/SaleTestHelper.java
```diff
@@ -0,0 +1,75 @@
...
+    userRequestDtoSeller.setId(sale.getSeller().getId());
+    dto.setSeller(userRequestDtoSeller);
+    userRequestDtoBuyer.setId(sale.getBuyer().getId());
+    dto.setBuyer(userRequestDtoBuyer);
+    dto.setTotalDiscountedPrice(sale.getProducts().get(0).getSalePrice());
```
VladoKat: if the sale has more than 1 product, this would be inaccurate. We should calculate it based on all the products

src/test/java/jewellery/inventory/helper/UserTestHelper.java
```diff
@@ -27,9 +27,19 @@ public static User createTestUser() {
     user.setRole(Role.ADMIN);
     return user;
   }
+  public static User createTestUserForSale() {
```
VladoKat: What makes this user eligible "for sale"? It seems like a regular user to me. FurtherMore it looks it does the same as  createTestUserWithId()

src/test/java/jewellery/inventory/helper/UserTestHelper.java
```diff
@@ -58,6 +68,20 @@ public static UserRequestDto createDifferentUserRequest() {
...
+    userRequest.setPassword(USER_PASSWORD);
+    return userRequest;
+  }
+  public static UserRequestDto createBuyerUserRequest() {
+    UserRequestDto userRequest = new UserRequestDto();
+    userRequest.setName("buyer");
+    userRequest.setEmail("buyer@example.com");
+    userRequest.setPassword(USER_PASSWORD);
+    return userRequest;
+  }
```
VladoKat: Why do we need those? Why can't we use the requests we've already created?

src/test/java/jewellery/inventory/integration/SaleIntegrationTest.java
```diff
@@ -0,0 +1,186 @@
...
+  private String getBaseSaleUrl() {
+    return getBaseUrl() + "/sales";
+  }
+
+  @Autowired private UserRepository userRepository;
+  @Autowired private SaleRepository saleRepository;
+  @Autowired private ProductRepository productRepository;
+  @Autowired private ResourceRepository resourceRepository;
+  @Autowired private ResourceInUserRepository resourceInUserRepository;
+  @Autowired private ResourceInProductRepository resourceInProductRepository;
```
VladoKat: We have these in common in almost every CRUD test. We should consider moving them to a base class. Let's not do it as part of this ticket but we should think about it in the next stages.

src/test/java/jewellery/inventory/integration/SaleIntegrationTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+  @Test
+  void getAllSalesSuccessfully() {
+
+    ResponseEntity<List<SaleResponseDto>> response =
+        this.testRestTemplate.exchange(
+            getBaseSaleUrl(), HttpMethod.GET, null, new ParameterizedTypeReference<>() {});
+
+    assertEquals(response.getStatusCode(), HttpStatusCode.valueOf(200));
+    assertNotNull(response.getBody());
```
VladoKat, line 108: I believe the response here would be an empty array because we don't have anything in the DB to fetch. Let's create at least one Sale and verify we've fetched it and it matches the structure we've agreed upon (List<SaleResponseDto>).

src/test/java/jewellery/inventory/integration/SaleIntegrationTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+    ResponseEntity<SaleResponseDto> saleResponse =
+        this.testRestTemplate.postForEntity(
+            getBaseSaleUrl(), saleRequestDto, SaleResponseDto.class);
+
+    assertEquals(HttpStatus.CREATED, saleResponse.getStatusCode());
+  }
+
+  @NotNull
+  private static SaleRequestDto getSaleRequestDto2(
```
VladoKat: Why is this named 2? It's never a good practice to append numbers to differentiate between similar methods/variables but in this case I don't even see another getSaleRequestDto2

src/test/java/jewellery/inventory/integration/SaleIntegrationTest.java
```diff
@@ -0,0 +1,186 @@
...
+        this.testRestTemplate.postForEntity(
+            getBaseProductUrl(), productRequestDto, ProductResponseDto.class);
+
+    SaleRequestDto saleRequestDto = getSaleRequestDto2(seller, buyer, productResponse);
+
+    ResponseEntity<SaleResponseDto> saleResponse =
+        this.testRestTemplate.postForEntity(
+            getBaseSaleUrl(), saleRequestDto, SaleResponseDto.class);
+
+    assertEquals(HttpStatus.CREATED, saleResponse.getStatusCode());
```
VladoKat, line 131: Please, add another meaningful assertion. We should verify the sale we've sent matches the details saved in the DB, meaning data that is returned in the body.

src/test/java/jewellery/inventory/integration/SaleIntegrationTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+  private void cleanAllRepositories() {
+    productRepository.deleteAll();
+    saleRepository.deleteAll();
+    userRepository.deleteAll();
+    productRepository.deleteAll();
+    resourceRepository.deleteAll();
+    resourceInUserRepository.deleteAll();
+    resourceInProductRepository.deleteAll();
+  }
```
VladoKat: I believe all of these are also used in multiple integration tests. I don't recommend refactoring it right now. Just mentally marking how much code can be extracted to the base class and just be reused after that by the specific tests.

src/test/java/jewellery/inventory/unit/mapper/SaleMapperTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class SaleMapperTest {
+  @Mock private SaleMapper saleMapper;
```
VladoKat: We are testing SaleMapper. We are not supposed to mock SaleMapper in this case. We should `@InjectMock` it and `@Mock` all of its dependencies. The methods of the mapper are supposed to be executed as they are here and mocking only the results of the dependencies it has.

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -258,7 +259,7 @@ void testDeleteProductShouldThrowExceptionWhenProductIsPartOfProduct() {
 
   @Test
   void testDeleteProductShouldThrowExceptionWhenProductIsSold() {
-    product.setSold(true);
+    product.setPartOfSale(new Sale()); //TODO
```
VladoKat: Is there something TODO here? If there is, 

If not, remove the comment.

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -282,7 +283,7 @@ void deleteProductShouldThrowExceptionWhenProductIsPartOfProduct() {
 
   @Test
   void deleteProductShouldThrowExceptionWhenProductIsSold() {
-    product.setSold(true);
+    product.setPartOfSale(new Sale()); //TODO
```
VladoKat: Same here.

src/test/java/jewellery/inventory/unit/service/SaleServiceTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+    when(saleRepository.findAll()).thenReturn(sales);
+
+    List<SaleResponseDto> responses = saleService.getAllSales();
+
+    Assertions.assertEquals(sales.size(), responses.size());
+  }
+
+  @Test
+  void testCreateSaleProductWhenDateIsCorrect() {
```
VladoKat: The test names should be following a specific pattern:
will<action>Successfully - happy path
will<action>Throws<Exception>When<condition> - unhappy path with reason.

I see in our code base that not all of them are following this pattern but we should start somewhere

src/test/java/jewellery/inventory/unit/service/SaleServiceTest.java
```diff
@@ -0,0 +1,122 @@
...
+    when(saleMapper.mapEntityToResponseDto(sale)).thenReturn(saleResponseDto);
+
+    SaleResponseDto actual = saleService.createSale(saleRequestDto);
+    assertNotEquals(actual.getBuyer(), actual.getSeller());
+    Assertions.assertEquals(saleRequestDto.getSellerId(), actual.getSeller().getId());
+    assertNotNull(actual);
+  }
+
+  @Test
+  void testCreateSaleProductWhenProductOwnerEqualsRecipientException() {
```
VladoKat: As mentioned above, this case should actually be allowed by our system.

src/test/java/jewellery/inventory/unit/service/SaleServiceTest.java
```diff
@@ -0,0 +1,122 @@
...
+            () -> saleService.createSale(saleRequestDtoOwnerEqualsRecipient));
+  }
+
+  @Test
+  void testProductOwnerNotSeller() {
+    UUID ownerId = UUID.randomUUID();
+    UUID sellerId = UUID.randomUUID();
+
+    try {
+      throw new ProductOwnerNotSeller(ownerId, sellerId);
```
VladoKat: We are supposed to stub the repository (or another related dependency) so that the service create method throws the exception. Right now we are just testing the constructor of the ProductOwnerNotSeller exception.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -94,7 +94,7 @@ private void throwExceptionIfProductOwnerEqualsRecipient(Product product, UUID r
   }
 
   private void throwExceptionIfProductIsSold(UUID id, Product product) {
-    if (product.isSold()) {
+    if (product.getPartOfSale()!=null) {
```
VladoKat: This doesn't look formatted.

VladoKat (changes requested):

src/main/java/jewellery/inventory/dto/request/ProductPriceDiscountRequestDto.java
berki0: if a field of type double (or other primitive numeric value) is null in Java, it will be converted to the value 0.0

VladoKat: Yeah but in this case does the `@NotNull` annotation do anything?

src/main/java/jewellery/inventory/exception/product/SellerNotOwnerException.java
```diff
@@ -2,8 +2,8 @@
 
 import java.util.UUID;
 
-public class UserNotOwnerException extends RuntimeException {
-    public UserNotOwnerException(UUID userId, UUID productId) {
+public class SellerNotOwnerException extends RuntimeException {
+    public SellerNotOwnerException(UUID userId, UUID productId) {
```
VladoKat: Hm... I think we renamed the wrong exception? Actually if we already have the UserNotOwnerException, I think we can reuse it instead of SellerNotOwner (ProductOwnerNotSeller)

src/main/java/jewellery/inventory/exception/product/ProductOwnerNotSeller.java
VladoKat, line 1: This is the one we should have renamed. I guess it can actually be removed after seeing we have a `UserNotOwnerException`

src/main/java/jewellery/inventory/mapper/ProductMapper.java
```diff
@@ -21,13 +21,16 @@ public ProductResponseDto mapToProductResponseDto(Product product) {
 
     ProductResponseDto productResponseDto = new ProductResponseDto();
     productResponseDto.setId(product.getId());
-    productResponseDto.setSold(product.isSold());
+    if (product.getPartOfSale()!=null){
```
VladoKat: This doesn't seem formatted.

src/main/java/jewellery/inventory/service/SaleService.java
VladoKat: `throwExceptionProductOwnerEqualsRecipientException` we don't want to check for this. We are okay if buyer and seller are the same user.

src/main/java/jewellery/inventory/service/SaleService.java
```diff
@@ -0,0 +1,92 @@
...
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class SaleService {
+  private final SaleRepository saleRepository;
+  private final SaleMapper saleMapper;
+  private final ProductRepository productRepository;
+  private final UserRepository userRepository;
```
VladoKat: I'd prefer if we used the UserService and ProductService for User and Products fetching/updating logic because the repos provide a very bare logic for dealing with the entities. We should be handling the entities through the services. You can consult with @ignatIgnatov about the complications involved and how to address them.

src/test/java/jewellery/inventory/helper/SaleTestHelper.java
VladoKat: The type is UserResponseDto but the variables are userRequestDtoSeller/Buyer. Perhaps should be renamed.
Still pending ^

src/test/java/jewellery/inventory/helper/UserTestHelper.java
VladoKat: What makes this user eligible "for sale"? It seems like a regular user to me. FurtherMore it looks it does the same as createTestUserWithId()
still pending^

src/test/java/jewellery/inventory/helper/UserTestHelper.java
VladoKat: Why do we need those? Why can't we use the requests we've already created?
still pending^

src/test/java/jewellery/inventory/integration/SaleIntegrationTest.java
VladoKat: ^

src/test/java/jewellery/inventory/integration/SaleIntegrationTest.java
VladoKat: ^

src/test/java/jewellery/inventory/integration/SaleIntegrationTest.java
VladoKat: ^

src/test/java/jewellery/inventory/unit/mapper/SaleMapperTest.java
VladoKat: ^

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
VladoKat: ^

VladoKat (changes requested):
Added some new comments. Please, take a look at the unresolved comments from before. I've marked them in this review as well

src/main/java/jewellery/inventory/exception/product/UserNotOwnerException.java
VladoKat, line 1: This is the exception we should use instead of ProductOwnerNotSeller. It is more generic and can be used for multiple cases. ProductOwnerNotSeller (or SellerNotOwner) would be a subset of it but they don't really give any more details.

Please, bring back this exception and use it instead of ProductOwnerNotSeller. Remove ProductOwnerNotSeller

src/main/java/jewellery/inventory/exception/product/ProductOwnerNotSeller.java
VladoKat: We don't need this exception if we have UserNotOwner.

src/main/java/jewellery/inventory/mapper/SaleMapper.java
```diff
@@ -0,0 +1,85 @@
...
+  private final ProductMapper productMapper;
+
+  public SaleResponseDto mapEntityToResponseDto(Sale sale) {
+    SaleResponseDto saleResponseDto = new SaleResponseDto();
+    saleResponseDto.setSeller(userMapper.toUserResponse(sale.getSeller()));
+    saleResponseDto.setBuyer(userMapper.toUserResponse(sale.getBuyer()));
+    saleResponseDto.setProducts(mapAllProductsToResponse(sale));
+    saleResponseDto.setTotalPrice(getTotalPriceFromEntity(sale.getProducts()));
+    saleResponseDto.setTotalDiscount(calculateDiscount(sale.getProducts(), "amount"));
+    saleResponseDto.setTotalDiscountedPrice(calculateDiscount(sale.getProducts(), "percentage"));
```
VladoKat: Please, extract "amount" and "percentage" into constants and use them instead of the hardcoded strings.

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -59,6 +57,16 @@ public ProductResponseDto getProduct(UUID id) {
     return productMapper.mapToProductResponseDto(product);
   }
 
+  public Product returnProductByID(UUID id) {
+    return productRepository.findById(id).orElseThrow(() -> new ProductNotFoundException(id));
+  }
```
VladoKat: We have a method named getProduct that has almost the same signature. Please, rename the other one to `getProductResponse` and rename this one to `getProduct`

src/main/java/jewellery/inventory/mapper/UserMapper.java
```diff
@@ -6,11 +6,12 @@
 import jewellery.inventory.model.User;
 import org.mapstruct.Mapper;
 
-@Mapper(
-    componentModel = "spring")
+@Mapper(componentModel = "spring")
 public interface UserMapper {
   UserRequestDto toUserRequest(User user);
 
+  User toUserEntity(UserResponseDto userResponseDto);
```
VladoKat: We shouldn't need this method. We can create a getUser method in UserService that returns the entity. In the service the current method that fetches the user is fetching an entity and mapping it to a ResponseDto. We can do the same as we have done for fetching the product

src/main/java/jewellery/inventory/dto/request/SaleRequestDto.java
VladoKat, line 1: I realize it's not mentioned in the ticket and I apologize for that but can we include the date in the request as well?

src/main/java/jewellery/inventory/dto/request/ProductPriceDiscountRequestDto.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.UUID;
+
+@Getter
+@Setter
+@RequiredArgsConstructor
+@AllArgsConstructor
+public class ProductPriceDiscountRequestDto {
+  @NotNull private UUID productId;
+   private double salePrice;
+   private double discount;
```
VladoKat: What happens if the user decides NOT to pass salePrice and discount in the request? Because the variables are double, it's going to assign them the default value - 0.0. I imagine we are going to set the price and discount of the product to 0. This shouldn't be the case. In this case we should not change the price and discount of the product. Let's make them `Double` and make checks if the values are not passed (they are null), then we wouldn't change them in the product.

src/test/java/jewellery/inventory/helper/UserTestHelper.java
VladoKat: Another try. I imagine you don't see this in the previous reviews because Github "hides" the comments if they are too many

src/test/java/jewellery/inventory/unit/mapper/SaleMapperTest.java
VladoKat, line 1: I believe file is not formatted, please, format it

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -138,20 +138,6 @@ void testCreateProductShouldThrowWhenProductNotFound() {
...
-    when(userRepository.findById(productRequestDto.getOwnerId())).thenReturn(Optional.of(user));
-    when(productRepository.findById(product.getId())).thenReturn(Optional.of(product));
-
-    User anotherUser = createTestUserWithRandomId();
-    product.setOwner(anotherUser);
-    productRequestDto.setProductsContent(List.of(product.getId()));
-
-    assertThrows(
-        UserNotOwnerException.class, () -> productService.createProduct(productRequestDto));
-  }
```
VladoKat, line 153: I don't think this test is obsolete. Why have we removed it?

VladoKat (changes requested):
Minor changes. We have misunderstood each other about the exceptions. Let's talk it through if needed.

src/main/java/jewellery/inventory/dto/request/SaleRequestDto.java
```diff
@@ -0,0 +1,30 @@
...
+@RequiredArgsConstructor
+@AllArgsConstructor
+public class SaleRequestDto {
+  @NotNull private UUID sellerId;
+  @NotNull private UUID buyerId;
+
+  @Valid @NotEmpty private List<ProductPriceDiscountRequestDto> products;
+
+  @NotNull
+  @Past(message = "Date must be in the past")
```
VladoKat: Let this be `@PastOrPresent`

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
VladoKat, line 1: There is no functional change to this file now. Can you rollback the changes? I recommend doing `git checkout origin/test src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java`

src/main/java/jewellery/inventory/mapper/UserMapper.java
VladoKat, line 1: This file has obsolete changes as well. Can you remove them? (`git checkout origin/test src/main/java/jewellery/inventory/mapper/UserMapper.java`)

src/main/java/jewellery/inventory/service/UserService.java
```diff
@@ -26,11 +26,16 @@ public List<UserResponseDto> getAllUsers() {
     return userMapper.toUserResponseList(userRepository.findAll());
   }
 
-  public UserResponseDto getUser(UUID id) {
+  public UserResponseDto getUserResponse(UUID id) {
     return userMapper.toUserResponse(
         userRepository.findById(id).orElseThrow(() -> new UserNotFoundException(id)));
   }
```
VladoKat:
```suggestion
  public UserResponseDto getUserResponse(UUID id) {
    return userMapper.toUserResponse(getUser(id));
  }
```

src/main/resources/db/migration/V1__insert_user_admin.sql
VladoKat, line 1: Why did we rename the migration names from ..1_1.. to ..1..? Was it on purpose?

src/test/java/jewellery/inventory/helper/UserTestHelper.java
VladoKat, line 1: File should be formatted.

src/test/java/jewellery/inventory/unit/service/SaleServiceTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    when(saleRepository.findAll()).thenReturn(sales);
+
+    List<SaleResponseDto> responses = saleService.getAllSales();
+
+    Assertions.assertEquals(sales.size(), responses.size());
+  }
+
+  @Test
+  void testCreateSaleProductWillSuccessfully() {
```
VladoKat:
```suggestion
  void testCreateSaleSuccessfully() {
```

src/test/java/jewellery/inventory/unit/service/SaleServiceTest.java
```diff
@@ -0,0 +1,111 @@
...
+    when(userService.getUser(any(UUID.class))).thenReturn(seller, buyer);
+    when(saleRepository.save(sale)).thenReturn(sale);
+    when(productService.getProduct(any(UUID.class))).thenReturn(product);
+
+    when(saleMapper.mapEntityToResponseDto(sale)).thenReturn(saleResponseDto);
+
+    SaleResponseDto actual = saleService.createSale(saleRequestDto);
+    assertNotEquals(actual.getBuyer(), actual.getSeller());
+    Assertions.assertEquals(saleRequestDto.getSellerId(), actual.getSeller().getId());
+    assertNotNull(actual);
```
VladoKat, line 96: You should either move this assertion above or remove it completely. Right now because of the previous assertions, if actual is null, it would throw a nullpointer before reaching this line.

VladoKat: We should make an assertion whether the product owner is the buyer

VladoKat (changes requested):
I think these should be the final final 🥵

src/main/java/jewellery/inventory/dto/request/SaleRequestDto.java
```diff
@@ -0,0 +1,29 @@
...
+@Setter
+@RequiredArgsConstructor
+@AllArgsConstructor
+public class SaleRequestDto {
+  @NotNull private UUID sellerId;
+  @NotNull private UUID buyerId;
+
+  @Valid @NotEmpty private List<ProductPriceDiscountRequestDto> products;
+
+  @PastOrPresent(message = "Date must be in the past")
```
VladoKat:
```suggestion
  @PastOrPresent(message = "Date must be in the past or present")
```

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -52,13 +50,23 @@ public List<ProductResponseDto> getByOwner(UUID ownerId) {
...
         productRepository.findById(id).orElseThrow(() -> new ProductNotFoundException(id));
 
     return productMapper.mapToProductResponseDto(product);
   }
 
+  public Product getProduct(UUID id) {
+    return productRepository.findById(id).orElseThrow(() -> new ProductNotFoundException(id));
+  }
+
+  public void updateProductOwner(Product product, User newOwner, Sale sale) {
```
VladoKat:
```suggestion
  public void updateProductOwnerAndSale(Product product, User newOwner, Sale sale) {
```

src/main/resources/db/migration/V2_1__update_resources.sql
VladoKat, line 1: If we are skipping major versions, we should be starting with 0 subversion -> `src/main/resources/db/migration/V2_0__update_resources.sql`

src/main/resources/db/migration/V3_1__change_fields_in_product.sql
VladoKat, line 1: Same here:
`src/main/resources/db/migration/V3_1__change_fields_in_product.sql`

VladoKat (changes requested):

VladoKat (approved):
Looks good to me! 🎖️

PR #57, opened 2023-11-22
Feature 38 sale product return

berki0:
adding methods for returning a sold product


src/test/java/jewellery/inventory/helper/ProductTestHelper.java
VladoKat, line 1: This file looks like it's not formatted.

src/main/java/jewellery/inventory/service/SaleService.java
VladoKat, line 1: We'd need to add a log annotation here for creating a system event. We might need to create a ProductReturnResponseDto just because of that reason. Let's discuss it when you can to see how to implement it exactly. I have an idea but would like to bounce it off you.

VladoKat (changes requested):
PR overall looks good. We need to find a way to also log an event when a product is returned.

src/test/java/jewellery/inventory/unit/service/SaleServiceTest.java
```diff
@@ -139,4 +151,74 @@ void testCreateSaleProductWillThrowsProductIsPartOfAnotherProduct() {
...
+    ProductReturnResponseDto productReturnResponseDto = saleService.returnProduct(product.getId());
+
+    assertEquals(0, sale.getProducts().size());
+    assertNull(productReturnResponseDto.getSaleAfter().getId());
+    assertNotNull(productReturnResponseDto.getReturnedProduct());
+    assertNull(productReturnResponseDto.getReturnedProduct().getPartOfSale());
+    assertNotEquals(
+        productReturnResponseDto.getReturnedProduct().getOwner().getId(),
+        productBeforeReturn.getOwner().getId());
+  }
```
VladoKat: if we have `testReturnProductSuccessfullyWithTwoProducts`, there is no need for this test I think.

VladoKat (changes requested):
Looks good overall. Apart from the one comment, I would try to extract some methods in the SaleCrudIntegration tests as I think there is repetition of code there and the tests are too long now.

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
VladoKat: This file is still not formatted.

VladoKat (approved):
Looks good

PR #62, opened 2023-11-25
Feature 43 database initial data

ignatIgnatov:


Alekseeybg:
I can not select file names to comment directly on them so I write here. Open for discussion question: Should not the migrations be in this order:
V1_1__init_schema.sql
V1_2__init_database.sql
V1_3__insert_user_admin.sql
V1_4__update_resources.sql
V1_5__change_fields_in_product.sql
With the note that V1_3__insert_user_admin.sql is actually not needed anymore since we can create the admin user in the V1_2__init_database.sql migration? Or it is not the way flyway works and we should not change anything?

Alekseeybg (approved):

PR #63, opened 2023-11-27
Feature 38 sale product return fix

berki0:


src/test/java/jewellery/inventory/integration/SaleCrudIntegrationTest.java
```diff
@@ -4,11 +4,11 @@
...
 import static jewellery.inventory.helper.UserTestHelper.*;
 import static jewellery.inventory.model.EventType.SALE_CREATE;
+import static jewellery.inventory.model.EventType.SALE_RETURN_PRODUCT;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNull;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
```
VladoKat: Is this right? I see we are using `org.junit.jupiter.api.Assertions.*` everywhere else. Why are we replacing it with `org.junit.Assert.` here?

VladoKat (changes requested):
The imports seem wrong to me. Let's discuss if you want

VladoKat (approved):
Looks good to me. @Alekseeybg do you want to also take a look before merging?

Alekseeybg (approved):

PR #64, opened 2023-11-29
Feature 42 user details extension

Alekseeybg:


src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
```diff
@@ -1,21 +1,34 @@
...
 @Data
 public class UserRequestDto {
   private static final String NAME_PATTERN_VALIDATION_MSG =
       "Name must only contain alphanumeric characters and underscores, and no consecutive underscores";
   private static final String PWD_PATTERN_VALIDATION_MSG =
       "Password must contain at least one digit, one lowercase letter, one uppercase letter, one special character, and be at least 8 characters long";
+  private static final String PHONE_PATTERN_VALIDATION_MSG =
+      "Phone number should be valid bulgarian mobile number";
+  private static final String NAME_REGEX = "^(?!.*__)[\\w]*$";
+  private static final String PHONE_REGEX = "^(\\+359|0)8[789]\\d(-)?\\d{3}(-)?\\d{3}$";
```
VladoKat: The number can be any number. We shouldn't be limited to Bulgarian only.

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
```diff
@@ -1,21 +1,34 @@
...
 
-  @NotBlank(message = "Name must not be blank, empty or null")
-  @Size(min = 3, max = 64, message = "Name size must be between 3 and 64")
-  @Pattern(regexp = "^(?!.*__)[A-Za-z0-9_]*$", message = NAME_PATTERN_VALIDATION_MSG)
-  private String name;
+  @NotBlank(message = "First name must not be blank, empty or null")
+  @Size(min = 3, max = 50, message = "First name must size must be between 3 and 64")
```
VladoKat: either make max to be 64 or change the message to say 50.

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
```diff
@@ -1,21 +1,34 @@
...
+  @Pattern(regexp = NAME_REGEX, message = NAME_PATTERN_VALIDATION_MSG)
+  private String firstName;
+
+  @NotBlank(message = "Last name must not be blank, empty or null")
+  @Size(min = 3, max = 50, message = "Last name size must be between 3 and 64")
```
VladoKat: same here

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
```diff
@@ -29,4 +42,23 @@ public class UserRequestDto {
...
+
+  @Pattern(regexp = PHONE_REGEX, message = PHONE_PATTERN_VALIDATION_MSG)
+  private String phone2;
+
+  @DateTimeFormat(pattern = "dd/MM/yyyy")
+  @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "dd/MM/yyyy")
+  @PastOrPresent(message = "Birth date must be a past or present date")
+  private LocalDate birthDate;
+
+  private String favouriteColor;
```
VladoKat: Turns out we don't need favouriteColor as well

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
```diff
@@ -29,4 +42,23 @@ public class UserRequestDto {
...
+
+  @Size(min = 3, max = 500, message = "Last name size must be between 3 and 64")
```
VladoKat: Change the message to state the maximum is 500.

src/main/java/jewellery/inventory/dto/response/UserResponseDto.java
VladoKat, line 1: We should have a mini version of the UserResponseDto that holds the id, first and last name, and the email. We would use it to show who the executor is in the SystemEvents.

src/main/java/jewellery/inventory/model/User.java
```diff
@@ -27,12 +28,18 @@ public class User implements UserDetails {
...
+  @Column private String lastName;
 
   @Column(unique = true)
   private String email;
 
+  @Column private String address;
+  @Column private String phone;
+  @Column private String phone2;
+  @Column private LocalDate birthDate;
+  @Column private String favouriteColor;
```
VladoKat: favouriteColor is not needed anymore.

src/main/java/jewellery/inventory/repository/UserRepository.java
```diff
@@ -8,7 +8,7 @@
 
 @Repository
 public interface UserRepository extends JpaRepository<User, UUID> {
-  Optional<User> findByName(String name);
+  Optional<User> findByFirstName(String name);
```
VladoKat: I don't know if findByName was used anywhere but it seems findByFirstName is not used anywhere now.

src/main/resources/db/migration/V1_1__insert_user_admin.sql
```diff
@@ -1,4 +1,4 @@
-INSERT INTO users (id, name, email, password, role)
+INSERT INTO users (id, first_name, email, password, role)
 VALUES ('88596531-7f0f-407d-b502-31833b8c8e8d', 'root', 'root@gmail.com', '$2a$12$fGuoN79WFwHPUmirHOlxIO9kdmMTBrlNGKob0ay4muxXNDePg38ri', 'ADMIN')
 ON CONFLICT (email) DO UPDATE
-SET name = 'root', email = 'root@gmail.com', password = '$2a$12$fGuoN79WFwHPUmirHOlxIO9kdmMTBrlNGKob0ay4muxXNDePg38ri', role = 'ADMIN';
\ No newline at end of file
+SET first_name = 'root', email = 'root@gmail.com', password = '$2a$12$fGuoN79WFwHPUmirHOlxIO9kdmMTBrlNGKob0ay4muxXNDePg38ri', role = 'ADMIN';
```
VladoKat: I think we can even skip the firstName of the root user.

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
```diff
@@ -211,7 +211,7 @@ void updateUserFailsWhenNameDuplicate() {
             requestUpdate,
             UserResponseDto.class);
 
-    assertEquals(HttpStatus.BAD_REQUEST, responseEntity.getStatusCode());
+    assertEquals(HttpStatus.OK, responseEntity.getStatusCode());
```
VladoKat: This is a bad "fix" for a test. The requirements have changed and this is no longer a valid test case altogether. We should remove it.

src/test/java/jewellery/inventory/unit/service/UserServiceTest.java
```diff
@@ -102,36 +97,22 @@ void createUserWhenEmailIsAlreadyTakenThenThrowDuplicateEmailException() {
...
-    when(userMapper.toUserEntity(userRequest)).thenReturn(user);
-
-    assertThrows(DuplicateNameException.class, () -> userService.createUser(userRequest));
-
-    verify(userRepository, never()).save(any(User.class));
-  }
-
   @Test
   @DisplayName("Should create a new user when the email and name are not taken")
   void createUserWhenEmailAndNameAreNotTaken() {
```
VladoKat: Please, change the name of the test to reflect the new requirements. We care if only email is taken now.

src/test/java/jewellery/inventory/unit/service/security/JwtTokenServiceTest.java
```diff
@@ -126,7 +126,7 @@ void isTokenValidWithExpiredTokenThrowsJwtExpiredException() {
   @Test
   void generateTokenWithNullUserDetailsThrowsIllegalArgumentException() {
     lenient().when(jwtUtils.getSigningKey()).thenReturn(key);
-    lenient().when(userDetails.getUsername()).thenReturn(USER_NAME);
+    lenient().when(userDetails.getUsername()).thenReturn(FIRST_NAME);
```
VladoKat: Considering username is no longer unique, shouldn't we be using the email as a username in the UserDetails service?

VladoKat (changes requested):
There are a lot of minor things to be fixed. The JwtTokenServiceTest made me realize that we were using the username to persist the user information in the security context. The first name of the user is no longer unique, so we should reconsider and change the Username when it comes to the security context. Email seems like the appropriate data to put in this case.

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
Alekseeybg: My bad. I did not intend to leave this 'fix' here. The test was supposed to be removed (already removed 3 other tests) this one I 'fixed' instead of commenting it out and then i forgot to remove it 🗡️

src/test/java/jewellery/inventory/unit/service/security/JwtTokenServiceTest.java
Alekseeybg: Yes, we are using the email, just the getter is getUsername. But it returns the email. I think I made it that way because User implements UserDetails where we have field Username and the @Getter annotation creates a getter and I Override it so it returns the email...
image

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
Alekseeybg: Removed the regex. It will accept any String

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
Alekseeybg: Fixed. Changed the message to say 50.

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
Alekseeybg: Fixed. Changed the message to say 50.

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
Alekseeybg: Did not know that. Removed.

src/main/java/jewellery/inventory/dto/request/UserRequestDto.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/dto/response/UserResponseDto.java
Alekseeybg: Maybe call it "ExecutorResponseDto" ?

src/main/java/jewellery/inventory/model/User.java
Alekseeybg: Removed.

src/main/java/jewellery/inventory/repository/UserRepository.java
Alekseeybg: Removed

src/main/resources/db/migration/V1_1__insert_user_admin.sql
Alekseeybg: Actually both firstName and lastName are required if I remember correctly? But this migration is obsolete since we talked about modifying the flyway migrations. After we merge his code onto test, I will merge test into my branch and will create a new migration for the User extension logic. The creation of the root user is handled by Ignat

src/test/java/jewellery/inventory/unit/service/UserServiceTest.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/integration/UserCrudIntegrationTest.java
Alekseeybg: Removed

src/main/resources/db/migration/V1_1__insert_user_admin.sql
Alekseeybg: They were not required in the version of the code you reviewed. Fixed.

src/test/java/jewellery/inventory/unit/service/UserServiceTest.java
```diff
@@ -102,36 +97,22 @@ void createUserWhenEmailIsAlreadyTakenThenThrowDuplicateEmailException() {
...
-    when(userRepository.findByName(user.getName())).thenReturn(Optional.of(user));
-    when(userMapper.toUserEntity(userRequest)).thenReturn(user);
-
-    assertThrows(DuplicateNameException.class, () -> userService.createUser(userRequest));
-
-    verify(userRepository, never()).save(any(User.class));
-  }
-
   @Test
   @DisplayName("Should create a new user when the email and name are not taken")
```
VladoKat: Can you also change the DisplayName?

VladoKat (changes requested):

src/test/java/jewellery/inventory/unit/service/security/JwtTokenServiceTest.java
VladoKat: Sure, I get that but in the test instead of email, we are stubbing it to return FIRST_NAME

src/test/java/jewellery/inventory/unit/service/UserServiceTest.java
Alekseeybg: Fixed

src/test/java/jewellery/inventory/unit/service/security/JwtTokenServiceTest.java
Alekseeybg: Fixed

src/main/java/jewellery/inventory/dto/response/UserResponseDto.java
Alekseeybg: Done

VladoKat (approved):

src/main/resources/application.properties
```diff
@@ -9,7 +9,7 @@ spring.flyway.password=${JMS_DATABASE_PASSWORD}
 spring.flyway.enabled=false
 
 # Specify the DB platform for Hibernate
-spring.jpa.hibernate.ddl-auto=update
+spring.jpa.hibernate.ddl-auto=none
```
VladoKat, line 12: Nice!

src/main/resources/db/migration/dev/V1_1__init_schema.sql
VladoKat, line 1: Why do we change this. From what I understand, we shouldn't change earlier versions of flyway migration.

VladoKat (changes requested):
Generally looks good. I am not approving because there is a problem with the update logic. There are also a few comments to be resolved.

src/main/resources/db/migration/dev/V1_1__init_schema.sql
Alekseeybg: Because it was missing a column, And more specifically the 'executor' column. And while I was at it I saw a little better way to write the query...

VladoKat: In this case I guess it wasn't working as expected before and service should have failed to start if we relied only on the script to create the DB?

Alekseeybg: I believe I noticed the problem after I tried to test with some request and it failed to create the event properly

Alekseeybg: It did not show up earlier because on the test branch hibernate was not disabled yet. So even if you delete the whole database (like i did) it creates the tables from the entities. After disabling hibernate and deletion of the database it will throw this error:
image

Alekseeybg:
I fixed the update logic. Should be ok now. It was fixed before but forgot to commit

VladoKat (changes requested):
Changes look good but we need to have a test for the password and role issue. In TDD, we'd create a test (or tests) for the bug we've found that would fail before the fix and then after writing the logic. the tests would pass.

TLDR: Let's create a test to make sure when updating a user, their password stays the same.

VladoKat (approved):
Looks good

VladoKat (approved):
Still looks good

PR #66, opened 2023-12-06
Feature 48 add top up cost

ignatIgnatov:


ignatIgnatov (commented):
Well done!

src/main/java/jewellery/inventory/dto/response/ResourcesInUserResponseDto.java
```diff
@@ -15,4 +15,5 @@
 public class ResourcesInUserResponseDto {
   private UserResponseDto owner;
   private List<ResourceQuantityResponseDto> resourcesAndQuantities;
+  private double dealPrice;
```
CodeLikeAlexito: I've noticed that in a lot of places in the project we use double and not BigDecimal datatype. In my opinion BigDecimal is better to be used:

https://www.linkedin.com/pulse/why-we-should-use-bigdecimal-instead-double-java-financial-ismail/
https://www.baeldung.com/java-double-vs-bigdecimal#:~:text=The%20BigDecimal%20class%20offers%20better,accurate%20than%20floating%2Dpoint%20arithmetic.

src/main/java/jewellery/inventory/service/ResourceInUserService.java
```diff
@@ -128,12 +129,23 @@ private ResourcesInUserResponseDto getResourceInUserResponse(UUID userId, UUID r
     return null;
   }
 
-  private ResourceInUser addResourceToUser(User user, Resource resource, Double quantity) {
-    ResourceInUser resourceInUser =
-        findResourceInUser(user, resource.getId())
-            .orElseGet(() -> createAndAddNewResourceInUser(user, resource, 0));
+  private void addResourceToUserToTransfer(User user, Resource resource, Double quantity) {
```
CodeLikeAlexito: Another place where BigDecimal would be good to be used. And plus, above we are using primitive double datatype and here the Double class. This is not very good approach.

src/main/java/jewellery/inventory/service/ResourceInUserService.java
```diff
@@ -128,12 +129,23 @@ private ResourcesInUserResponseDto getResourceInUserResponse(UUID userId, UUID r
...
+    ResourceInUser resourceInUser = getResourceInUser(user, resource);
+    resourceInUser.setQuantity(resourceInUser.getQuantity() + quantity);
+    resourceInUserRepository.save(resourceInUser);
```
VladoKat: This method isn't very different than the one that just adds resource to user when doing top-up (a.k.a. delivery). We should consider reusing the existing logic and just extend it for the top-up logic.

src/test/java/jewellery/inventory/helper/UserTestHelper.java
```diff
@@ -76,7 +76,17 @@ public static UserRequestDto createInvalidUserRequest() {
...
 
   public static @NotNull ResourceInUserRequestDto createResourceInUserRequestDto(
-      UUID userId, UUID resourceId, double quantity) {
+      UUID userId, UUID resourceId, double quantity, Double price) {
+    ResourceInUserRequestDto requestDto = new ResourceInUserRequestDto();
+    requestDto.setUserId(userId);
+    requestDto.setResourceId(resourceId);
+    requestDto.setQuantity(quantity);
+    requestDto.setDealPrice(price);
+    return requestDto;
```
VladoKat:
```suggestion
    ResourceInUserRequestDto requestDto = createResourceInUserRequestDtoWithoutPrice(userId, resourceId, quantity);
    requestDto.setDealPrice(price);
    return requestDto;
```

VladoKat (changes requested):
I left comments for a few small changes to the existing implementation. There is something more major I'd like to change though. With the current implementation we'd save the dealPrice in every ResourcesInUserResponse. This means it'd show in the removeQuantity and transferQuantity system events as well. We should make it so that it doesn't show dealPrice in these scenarios. Let's discuss about an approach if you want to.

src/main/java/jewellery/inventory/dto/request/ResourceInUserRequestDto.java
VladoKat, line 1: This seems like an obsolete change?

src/main/java/jewellery/inventory/dto/request/ResourcePurchaseRequestDto.java
VladoKat, line 1: How about if we made this to extend the ResourceInUserRequestDto? Then it'd just have the dealPrice as a new field. The rest I see are copied from there anyway

src/main/java/jewellery/inventory/dto/response/ResourcePurchaseResponseDto.java
VladoKat, line 1: Same with this. Can it extend ResourceInUserResponseDto?

VladoKat (changes requested):
I think if we make the new classes for purchase to extend the existing ResourceInUser model/dto, we'd not need to make some of the additional methods we've made in the ResourceInUser service.

src/main/java/jewellery/inventory/dto/request/ResourcePurchaseRequestDto.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/dto/response/ResourcePurchaseResponseDto.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/ResourceInUserService.java
```diff
@@ -96,8 +96,11 @@ public ResourcesInUserResponseDto removeQuantityFromResourceNoLog(
     User user = findUserById(userId);
     ResourceInUser resourceInUser = findResourceInUserOrThrow(user, resourceId);
 
-    return resourcesInUserMapper.toResourcesInUserResponseDto(
-        removeQuantityFromResource(resourceInUser, quantity));
+    if (resourceInUser != null) {
+      return resourcesInUserMapper.toResourcesInUserResponseDto(
+          removeQuantityFromResource(resourceInUser, quantity));
+    }
```
VladoKat: Is this null check necessary? From what I see on line 97 (findResourceInUser**OrThrow**) resourceInUser can't be null.

VladoKat (changes requested):
Last minor change/question. It looks great otherwise.

src/main/java/jewellery/inventory/service/ResourceInUserService.java
ignatIgnatov: I forgot to put the removeQuantityFromResource method before the if-statement. This method can return null value of ResourceInUser. Now I fixed it.

VladoKat (approved):
LGTM

PR #71, opened 2023-12-15
Remove resource quantity validation from service

Alekseeybg:
Refactored the code so we use controller level validation for quantity provided by annotation.


VladoKat (approved):
Looks good!

ignatIgnatov (approved):
I approve!

PR #72, opened 2023-12-18
Feature 53 expose edit for product

ignatIgnatov:


src/main/java/jewellery/inventory/dto/request/ResourceInUserRequestDto.java
```diff
@@ -12,14 +12,14 @@
...
+  private static final String QUANTITY_MIN_VALUE_MSG = "Quantity should not be less than 0.01.";
   private static final String QUANTITY_DECIMAL_PLACES_MSG =
-      "Quantity should not have more than 2 decimal places.";
+          "Quantity should not have more than 2 decimal places.";
 
   @NotNull private UUID userId;
   @NotNull private UUID resourceId;
 
-  @Min(value = 0, message = QUANTITY_MIN_VALUE_MSG)
+  @DecimalMin(value = "0.01", message = QUANTITY_MIN_VALUE_MSG)
```
berki0: it might be good to declare it in a constant as in "ResourceQuantityRequestDto"

ignatIgnatov: Done!

src/main/java/jewellery/inventory/dto/request/ProductRequestDto.java
```diff
@@ -1,24 +1,24 @@
...
 @RequiredArgsConstructor
 @AllArgsConstructor
 public class ProductRequestDto {
-
   @NotEmpty
   private List<UUID> authors;
   @NotNull
   private UUID ownerId;
   @NotEmpty
+  @Valid
```
VladoKat, line 21: I was wondering whether this is needed. If I understand the [following article](https://www.baeldung.com/spring-valid-vs-validated#using-valid-annotation-to-mark-nested-objects) right, turns out we need to. Now I am wondering whether we have other fields within request dtos that would need to be annotated as well.

src/main/java/jewellery/inventory/dto/request/ResourceInUserRequestDto.java
```diff
@@ -12,14 +12,15 @@
...
+  private static final String QUANTITY_MIN_VALUE_MSG = "Quantity should not be less than 0.01.";
   private static final String QUANTITY_DECIMAL_PLACES_MSG =
-      "Quantity should not have more than 2 decimal places.";
+          "Quantity should not have more than 2 decimal places.";
 
   @NotNull private UUID userId;
   @NotNull private UUID resourceId;
 
-  @Min(value = 0, message = QUANTITY_MIN_VALUE_MSG)
+  @DecimalMin(value = MINIMAL_RESOURCE_QUANTITY, message = QUANTITY_MIN_VALUE_MSG)
```
VladoKat: Do we need to change this? I think if that's the case, maybe we need to update it in other double values within our requests.

src/main/java/jewellery/inventory/exception/GlobalExceptionHandler.java
```diff
@@ -60,7 +62,9 @@ public ResponseEntity<Object> handleTypeMismatch(MethodArgumentTypeMismatchExcep
   public ResponseEntity<Object> handleBadDataExceptions(RuntimeException ex) {
     String errorMessage = ex.getMessage();
     if (ex instanceof ConstraintViolationException cve) {
-      errorMessage = cve.getConstraintViolations().iterator().next().getMessage();
+      errorMessage = cve.getConstraintViolations().stream()
+              .map(ConstraintViolation::getMessage)
+              .collect(Collectors.joining("; "));
```
VladoKat: This will affect the frontend messages in some cases. I think it's good but we should also think of some way to make our errorMessages mappable to messages to show the User in the frontend. This would not be part of this ticket, of course. Just thinking;

src/main/java/jewellery/inventory/dto/request/resource/ResourceQuantityRequestDto.java
```diff
@@ -1,16 +1,24 @@
...
 
+    private static final String MINIMAL_RESOURCE_QUANTITY = "0.01";
+    private static final String RESOURCE_QUANTITY_MESSAGE = "Resource quantity should not be less than 0.01";
+    private static final String QUANTITY_DECIMAL_PLACES_MSG =
+            "Resource quantity should not have more than 2 decimal places.";
+
     private UUID id;
+    @DecimalMin(value = MINIMAL_RESOURCE_QUANTITY, message = RESOURCE_QUANTITY_MESSAGE)
+    @Digits(integer = 10, fraction = 2, message = QUANTITY_DECIMAL_PLACES_MSG)
     private double quantity;
```
VladoKat: I am wondering whether we can create a model named something like EmbeddableResourceQuantity that would hold just the quantity field with all its constraints. This way we will avoid writing the same constraints in multiple entities. You can check this [stack overflow question](https://stackoverflow.com/a/72740830/13062206). Give it a try and let me know

src/main/resources/db/migration/dev/V1_1__drop_existed_schema.sql
VladoKat, line 1: These are the tables I have in postgre:
Screenshot 2023-12-22 at 9 14 44
Do we need to delete flyway_schema_history and product_author as well?

src/main/resources/db/migration/dev/V1_6__update_system_event_type_check_constraint.sql
VladoKat, line 1: We were wondering with @Alekseeybg whether it makes sense to have a constraint for the system event type at all.  The arguments against are:
1. System events are not created by user input. They are automatically created when something meaningful happens in the app. That means we aren't able to insert something that shouldn't be there.
2. Next time we create another system event type, we'd need to recreate this constraint

The arguments for are:
1. Someone with access to the DB could manually add a misstyped event. 

The second list is a bit thin in my opinion. Any thoughts/research on that?

VladoKat (changes requested):
Changes overall look good. Good job! I left some comments. Most of them are questions out of curiosity and possible future improvements. There are some little changes that need to be made or investigated.

VladoKat (approved):
Looks good!

PR #75, opened 2024-01-03
Refactor 55 use big decimal instead of double

Alekseeybg:


src/main/java/jewellery/inventory/dto/UserQuantityDto.java
VladoKat, line 1: This one has been bugging me for quite some time. Let's rename it to UserQuantityResponseDto and move it to the response folder. This PR is touching everywhere, so this seems like an opportune time to do this change.

src/main/java/jewellery/inventory/dto/request/resource/PearlRequestDto.java
VladoKat, line 1: Let's add some validation to the resource requests as well. At least the ones that are common sense: size, dimensions, carat, pricePerQuantity should all be positive. (writing this here to not spam everywhere in the other files)

src/main/java/jewellery/inventory/service/ResourceInUserService.java
```diff
@@ -142,32 +143,32 @@ private ResourcesInUserResponseDto getResourceInUserResponse(UUID userId, UUID r
...
-      ResourceInUser resourceInUser, double quantityToRemove) {
+      ResourceInUser resourceInUser, BigDecimal quantityToRemove) {
 
-    double totalQuantity = resourceInUser.getQuantity();
-    double newQuantity = totalQuantity - quantityToRemove;
+    BigDecimal totalQuantity = resourceInUser.getQuantity();
+    BigDecimal newQuantity = totalQuantity.subtract(quantityToRemove);
 
-    if (newQuantity < 0) {
+    if (newQuantity.compareTo(new BigDecimal("0")) < 0) {
```
VladoKat: These compareTo conditions are a bit hard to understand. Should we extract them to methods? Name them smth like `lessThanZero` and `approachesZero` or.. `isEffectivelyZero` for the other

src/main/java/jewellery/inventory/service/SystemEventService.java
```diff
@@ -44,6 +45,7 @@ public <T> void logEvent(EventType type, T entity) {
   }
 
   private <U> Object createMap(U entity) {
+    objectMapper.configure(JsonGenerator.Feature.WRITE_BIGDECIMAL_AS_PLAIN, true);
```
VladoKat, line 48: Should this configuration of the ObjectMapper be done in a @ Configuration class? I see we are only using it here for now and it may turn out to be a overkill. What do you think?

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
```diff
@@ -84,8 +90,11 @@ void addResourceToUserFailsWhenUserNotFound() {
 
     ResponseEntity<ResourcesInUserResponseDto> response =
         sendAddResourceInUserRequest(
-            createResourceInUserRequestDto(
-                nonExistentUserId, createdResource.getId(), RESOURCE_QUANTITY));
+            createResourcePurchaseRequestDto(
+                nonExistentUserId,
+                createdResource.getId(),
+                RESOURCE_QUANTITY,
+                new BigDecimal("105.5").setScale(2, RoundingMode.HALF_UP)));
```
VladoKat: Repetition of this 105.5 value? Perhaps a constant should be introduced

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
VladoKat, line 1: There are some repetition of big decimals here. It may be worth extracting them in constants.

src/test/java/jewellery/inventory/integration/SaleCrudIntegrationTest.java
VladoKat, line 1: This file doesn't seem formatted.

src/test/java/jewellery/inventory/integration/SaleCrudIntegrationTest.java
```diff
@@ -222,12 +227,13 @@ private static SaleRequestDto getSaleRequestDto(
   }
 
   @NotNull
-  private static ResourceInUserRequestDto getResourceInUserRequestDto(
+  private static ResourcePurchaseRequestDto getResourceInUserRequestDto(
       User user, PreciousStone preciousStone) {
-    ResourceInUserRequestDto resourceInUserRequestDto = new ResourceInUserRequestDto();
+    ResourcePurchaseRequestDto resourceInUserRequestDto = new ResourcePurchaseRequestDto();
```
VladoKat: rename also the variable, not just the types?

VladoKat (changes requested):
Looks good overall. Take a look at the comments. There is one major improvement I think can be made: Let's have a BigDecimalUtil file that hold static methods for creating instances of BigDecimals (e.g. `new BigDecimal("20").setScale(2, RoundingMode.HALF_UP)` -> getBigDecimal("20")). Then all the places that we are using BigDecimals would have a bit less code and therefore would be easier to read?

src/main/java/jewellery/inventory/service/SystemEventService.java
Alekseeybg: It is just for the objectMapper we are using in the tests so maybe its overkill

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
```diff
@@ -10,20 +14,17 @@
...
-
 public class ProductTestHelper {
 
   public static ResourceInUser getResourceInUser(User user, Resource pearl) {
     ResourceInUser resourceInUser = new ResourceInUser();
     resourceInUser.setId(UUID.randomUUID());
     resourceInUser.setOwner(user);
     resourceInUser.setResource(pearl);
-    resourceInUser.setQuantity(20);
+    resourceInUser.setQuantity(BigDecimalUtil.getBigDecimal("20"));
```
VladoKat: We can import this statically to keep the call a bit short - will result only in getBigDecimal("20") call

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
```diff
@@ -91,19 +92,19 @@ public static ProductResponseDto getReturnedProductResponseDto(
     productResponseDto.setProductsContent(null);
     productResponseDto.setContentOf(null);
     productResponseDto.setDescription("description");
-    productResponseDto.setSalePrice(1000);
+    productResponseDto.setSalePrice(BigDecimalUtil.getBigDecimal("1000"));
     productResponseDto.setPartOfSale(null);
     productResponseDto.setCatalogNumber("Catalog Number");
     productResponseDto.setProductionNumber("Production Number");
-    productResponseDto.setDiscount(0);
+    productResponseDto.setDiscount(new BigDecimal("0"));
```
VladoKat: I noticed we don't use the BigDecimalUtil.getBigDecimal method for 0 values. Is there any reason for that?

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
```diff
@@ -110,7 +118,11 @@ void addResourceToUserFailsWhenQuantityInvalid() {
 
     ResponseEntity<ResourcesInUserResponseDto> response =
         sendAddResourceInUserRequest(
-            createResourceInUserRequestDto(createdUser.getId(), createdResource.getId(), -5));
+            createResourcePurchaseRequestDto(
+                createdUser.getId(),
+                createdResource.getId(),
+                new BigDecimal("-5"),
+                new BigDecimal("-105.5")));
```
VladoKat: How about taking advantage of the BigDecimal::negate() ?

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
```diff
@@ -237,18 +255,18 @@ void removeResourceQuantityFromUserSuccessfully() throws JsonProcessingException
...
         sendDeleteQuantityFromResourceInUserRequest(
-            createdUser.getId(), createdResource.getId(), 1.0);
+            createdUser.getId(), createdResource.getId(), RESOURCE_QUANTITY_TO_REMOVE);
 
     ResponseEntity<ResourcesInUserResponseDto> resourcesInUserResponse =
         sendGetResourcesInUserRequest(createdUser.getId());
     ResourceQuantityResponseDto resourceQuantity =
         findResourceQuantityIn(createdResource.getId(), resourcesInUserResponse);
-    assertEquals(RESOURCE_QUANTITY - 1, resourceQuantity.getQuantity(), 0.01);
+    assertEquals(RESOURCE_QUANTITY.subtract(new BigDecimal("1")), resourceQuantity.getQuantity());
```
VladoKat:
```suggestion
    assertEquals(RESOURCE_QUANTITY.subtract(QUANTITY_TO_REMOVE), resourceQuantity.getQuantity());
```

VladoKat (changes requested):

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
Alekseeybg: The regex I used to find/replace for some reason did not catch BigDecimal("0"). I have no explanation why. Will fix.

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
Alekseeybg: Good suggestion! I did not know about negate()

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
Alekseeybg: Fixed but instead of using the utility method I used the recommended way - BigDecimal.ZERO. That is specifically for where we had "new BigDecimal("0")".

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
```diff
@@ -133,7 +142,9 @@ void updateResourceInUserSuccessfullyWhenAddedMultipleTimes() {
         response.getBody().getResourcesAndQuantities();
     assertNotNull(resourceQuantities);
     assertEquals(1, resourceQuantities.size());
-    assertEquals(RESOURCE_QUANTITY * 2, resourceQuantities.get(0).getQuantity(), 0.001);
+    assertEquals(
+        RESOURCE_QUANTITY.multiply(new BigDecimal("2")).setScale(2, RoundingMode.HALF_UP),
```
VladoKat: Another new BigDecimal. Should we try to get rid of all of these and just create them from the util class?

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
```diff
@@ -285,12 +301,14 @@ void removeResourceQuantityFromUserFailsWhenInsufficientQuantity() {
...
-            createdUser.getId(), createdResource.getId(), RESOURCE_QUANTITY));
+        createResourcePurchaseRequestDto(
+            createdUser.getId(), createdResource.getId(), RESOURCE_QUANTITY, RESOURCE_PRICE));
 
     ResponseEntity<ResourcesInUserResponseDto> response =
         sendDeleteQuantityFromResourceInUserRequest(
-            createdUser.getId(), createdResource.getId(), RESOURCE_QUANTITY + 1);
+            createdUser.getId(),
+            createdResource.getId(),
+            RESOURCE_QUANTITY.add(BigDecimalUtil.getBigDecimal("1")));
```
VladoKat: static import?

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
```diff
@@ -300,12 +318,12 @@ void removeResourceQuantityFromUserFailsWhenQuantityNegative() {
...
     sendAddResourceInUserRequest(
-        createResourceInUserRequestDto(
-            createdUser.getId(), createdResource.getId(), RESOURCE_QUANTITY));
+        createResourcePurchaseRequestDto(
+            createdUser.getId(), createdResource.getId(), RESOURCE_QUANTITY, RESOURCE_PRICE));
 
     ResponseEntity<ResourcesInUserResponseDto> response =
         sendDeleteQuantityFromResourceInUserRequest(
-            createdUser.getId(), createdResource.getId(), -1.0);
+            createdUser.getId(), createdResource.getId(), BigDecimal.valueOf(-1.0));
```
VladoKat: should we use the getBigDecimal("-1")?

src/test/java/jewellery/inventory/unit/mapper/SaleMapperTest.java
```diff
@@ -53,7 +55,10 @@ void setUp() {
     productsForSale = SaleTestHelper.getProductsList(product);
     sale = SaleTestHelper.createSaleWithTodayDate(seller, buyer, productsForSale);
     productPriceDiscountRequestDto =
-        SaleTestHelper.createProductPriceDiscountRequest(product.getId(), 1000, 10);
+        SaleTestHelper.createProductPriceDiscountRequest(
+            product.getId(),
+            BigDecimalUtil.getBigDecimal("1000"),
```
VladoKat: static import?

VladoKat (changes requested):

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
Alekseeybg: There were quite a few more places that were not using static import.. I think I found them all and fixed them.

src/test/java/jewellery/inventory/integration/ResourceInUserCrudIntegrationTest.java
Alekseeybg: I found them all and fixed all but one. In the object mapper config in the SystemEventTestHelper for some reason if I change it to use the utility method there is a test that starts failing...

VladoKat (approved):

PR #76, opened 2024-01-07
clear ResourceInProductRepository when ResourceInProduct quantity is zero

ignatIgnatov:
clear ResourceInProductRepository when ResourceInProduct quantity is zero by product delete or product update operation

### DIFF
--- a/src/main/java/jewellery/inventory/service/ProductService.java
+++ b/src/main/java/jewellery/inventory/service/ProductService.java
@@ -72,7 +72,7 @@ public class ProductService implements EntityFetcher {
   }
 
   public ProductReturnResponseDto getProductReturnResponseDto(
-          SaleResponseDto sale, Product product) {
+      SaleResponseDto sale, Product product) {
     return productMapper.mapToProductReturnResponseDto(sale, product);
   }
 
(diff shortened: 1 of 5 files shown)


berki0 (approved):

src/main/java/jewellery/inventory/service/ProductService.java
```diff
@@ -198,20 +200,16 @@ private void disassembleProductContent(Product product) {
...
-  }
-
-  private void moveQuantityFromResourcesInProductToResourcesInUser(
-      List<ResourceInProduct> resourcesInProduct, User owner) {
     resourcesInProduct.forEach(
         resourceInProduct -> {
           resourceInUserService.addResourceToUserNoLog(
-              getResourceInUserRequest(owner, resourceInProduct));
+              getResourceInUserRequest(product.getOwner(), resourceInProduct));
           resourceInProduct.setQuantity(0);
```
VladoKat: How about instead of `resourceInProduct.setQuantity(0)`, we just do `resourceInProductRepository.delete(resourceInProduct.getId())`? In this case we don't need to `clearResourceInProductRepository` that fetches all of the records in ResourceInProduct table just to remove the ones that are with 0 quantity.

VladoKat (changes requested):

VladoKat (approved):
looks good!

VladoKat (approved):

PR #80, opened 2024-01-23
[bugfix-64-user-validation-fix] User validation fix

Alekseeybg:
User firstName and lastName accept cyrillic characters. Password does not require special symbol anymore.


ignatIgnatov (approved):
Looks good!

PR #83, opened 2024-02-04
Feature 65 expose organization create/read endpoint

CodeLikeAlexito:
Initial work for the new epic Organizations

Changes:

OrganizationController
OrganizationService
OrganizationRequestDto
OrganizationResponseDto

...


src/main/java/jewellery/inventory/dto/response/OrganizationResponseDto.java
```diff
@@ -0,0 +1,20 @@
...
+@AllArgsConstructor
+public class OrganizationResponseDto {
+  private UUID id;
+  private String name;
+  private String address;
+  private String note;
+  private List<ResourceInOrganizationResponseDto> resourceInOrganization;
+  private List<ProductResponseDto> productsOwned;
+  private List<UserInOrganizationResponseDto> userInOrganizations;
+  private List<SaleResponseDto> sales;
```
VladoKat: According to the AC we should only have id, name, address and note in the response. Please remove the rest of the dtos from the response.

src/main/java/jewellery/inventory/dto/response/ResourceInOrganizationResponseDto.java
VladoKat, line 1: As per the previous comment, this dto shouldn't be needed.

src/main/java/jewellery/inventory/dto/response/UserInOrganizationResponseDto.java
VladoKat, line 1: As per the previous comment, this dto shouldn't be needed.

src/main/java/jewellery/inventory/model/Organization.java
```diff
@@ -0,0 +1,40 @@
...
+  private String address;
+  @Column
+  private String note;
+  @OneToMany(mappedBy = "organization", cascade = CascadeType.ALL)
+  private List<ResourceInOrganization> resourceInOrganization;
+  @OneToMany(mappedBy = "organization", cascade = CascadeType.ALL)
+  private List<Product> productsOwned;
+  @OneToMany(mappedBy = "organization", cascade = CascadeType.ALL)
+  @ToString.Exclude
+  private List<UserInOrganization> userInOrganizations;
```
VladoKat:
```suggestion
  private List<UserInOrganization> usersInOrganization;
```

src/main/java/jewellery/inventory/model/ResourceInOrganization.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @ManyToOne
+  private Organization organization;
+
+  @ManyToOne private Resource resource;
+
+  private BigDecimal quantity;
+
+  @Transient
+  private BigDecimal dealPrice;
```
VladoKat, line 40: This dealPrice shouldn't be in the entity. We can ignore it for now as I saw we are doing the same as in the ResourceInUser but I think we should change it when we start implementing the new logic for adding resources to organizations.

src/main/java/jewellery/inventory/model/User.java
```diff
@@ -26,7 +26,7 @@ public class User implements UserDetails {
   private UUID id;
 
   @ManyToMany(mappedBy = "authors")
-  private List<Product> authoredProducts;
+  private transient List<Product> authoredProducts;
```
VladoKat: why do we need to have this as transient?

src/main/java/jewellery/inventory/model/User.java
VladoKat, line 1: This file seems as if it was experimented with and then wasn't brought back to its original state. It needs to be formatted as well.

src/main/java/jewellery/inventory/model/User.java
```diff
@@ -49,13 +51,17 @@ public class User implements UserDetails {
   private Role role = Role.USER;
 
   @OneToMany(mappedBy = "owner", cascade = CascadeType.ALL)
-  private List<Product> productsOwned = new ArrayList<>();
+  private  List<Product> productsOwned = new ArrayList<>();
 
   @OneToMany(mappedBy = "owner", cascade = CascadeType.ALL, orphanRemoval = true)
-  private List<ResourceInUser> resourcesOwned = new ArrayList<>();
+  @ToString.Exclude
```
VladoKat: Why do we exclude only the ResourcesInUser?

src/main/java/jewellery/inventory/service/OrganizationService.java
```diff
@@ -0,0 +1,68 @@
...
+    logger.debug("Fetching all organizations");
+    return organizationRepository.findAll();
+  }
+
+  private Organization getOrganization(UUID id) {
+    logger.info("Get organization by ID: {}", id);
+    return organizationRepository
+        .findById(id)
+        .orElseThrow(() -> new OrganizationNotFoundException(id));
+  }
```
VladoKat: Let's move the private methods below the public ones.

src/main/java/jewellery/inventory/service/OrganizationService.java
```diff
@@ -0,0 +1,68 @@
...
+  private final AuthService authService;
+  private final UserService userService;
+
+  private List<Organization> getAll() {
+    logger.debug("Fetching all organizations");
+    return organizationRepository.findAll();
+  }
+
+  private Organization getOrganization(UUID id) {
+    logger.info("Get organization by ID: {}", id);
```
VladoKat: What was our strategy for logging? I think fetching data should be debug and altering data in db should be info. Also, I see we are using this method in `getOrganizationResponse` and we are loggin the same there. Perhaps remove it from here and leave it only there?

src/main/java/jewellery/inventory/service/OrganizationService.java
```diff
@@ -0,0 +1,68 @@
...
+  private UserInOrganization getUserInOrganizationOwner(Organization organization) {
+    UserInOrganization userInOrganizationOwner = new UserInOrganization();
+    User user = userService.getUser(authService.getCurrentUser().getId());
+    userInOrganizationOwner.setUser(user);
+    userInOrganizationOwner.setOrganization(organization);
+    userInOrganizationOwner.setOrganizationPermission(
+        Arrays.asList(OrganizationPermission.values()));
+    logger.info("Created UserInOrganization for Organization ID: {}", organization.getId());
+    return userInOrganizationOwner;
+  }
```
VladoKat: We can add `organization.setUserInOrganizations(List.of(getUserInOrganizationOwner(organization)));` in this method, make it void and rename it to `makeCurrentUserOwner(organization)`. It seems a bit more descriptive to me this way.

src/main/resources/db/migration/dev/V1_7__create_organization_table.sql
VladoKat, line 1: Why are we inserting this migration before the product and resource sale ones?

src/main/resources/db/migration/dev/V1_7__create_organization_table.sql
```diff
@@ -0,0 +1,32 @@
...
+CREATE TABLE public.user_in_organization (
+    id UUID PRIMARY KEY,
+    user_id UUID,
+    organization_id UUID,
+    organization_permission VARCHAR(255) ARRAY,
+    CONSTRAINT fk_user_user_in_org FOREIGN KEY (user_id) REFERENCES users(id),
+    CONSTRAINT fk_org_user_in_org FOREIGN KEY (organization_id) REFERENCES organizations(id)
+);
+
+CREATE TABLE resource_in_organization (
```
VladoKat: what is the difference between public.<tableName> and just <tableName>? Does it make any difference?

src/main/resources/log4j2.properties
```diff
@@ -2,7 +2,7 @@
 appender.console.type = Console
 appender.console.name = ConsoleAppender
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = [%-5level] %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %c{1} - %msg%ex{5}%n
+appender.console.layout.pattern = [%-5level] %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %c{1} - %msg%ex%n
```
VladoKat, line 5: Nice!

src/test/java/jewellery/inventory/helper/OrganizationTestHelper.java
```diff
@@ -0,0 +1,64 @@
...
+    organization.setNote("Test Note");
+    return organization;
+  }
+
+  public static Organization getTestOrganizationWithUserInOrganizations() {
+    Organization organization = new Organization();
+    organization.setId(UUID.randomUUID());
+    organization.setName("Test Name");
+    organization.setAddress("Test Address");
+    organization.setNote("Test Note");
```
VladoKat: Pls extract these magic strings as constants if we are using them in multiple methods.

src/test/java/jewellery/inventory/helper/UserTestHelper.java
VladoKat, line 1: Format the file?

src/test/java/jewellery/inventory/integration/OrganizationCrudIntegrationTest.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @Test
+  void getOrganizationsSuccessfully() {
+    ResponseEntity<List<ProductResponseDto>> response =
+        this.testRestTemplate.exchange(
+            getBaseOrganizationsUrl(), HttpMethod.GET, null, new ParameterizedTypeReference<>() {});
+
+    assertEquals(response.getStatusCode(), HttpStatusCode.valueOf(200));
+    assertNotNull(response.getBody());
```
VladoKat, line 47: This is not a meaningful assertion. I know we are doing a lot of this kind of assertions in other tests but we should be more specific with our assertions. We should rework our tests setup. Ignore this for now. We will do it in a separate ticket

VladoKat (changes requested):

src/main/java/jewellery/inventory/service/OrganizationService.java
```diff
@@ -0,0 +1,66 @@
...
+  private final AuthService authService;
+  private final UserService userService;
+
+  public List<OrganizationResponseDto> getAllOrganizationsResponses() {
+    logger.debug("Fetching all organizationsResponses");
+    return getAll().stream().map(organizationMapper::toResponse).toList();
+  }
+
+  public OrganizationResponseDto getOrganizationResponse(UUID id) {
+    logger.info("Get organizationResponse by ID: {}", id);
```
VladoKat: Let's make this a debug log

VladoKat (changes requested):

VladoKat (approved):

PR #84, opened 2024-02-14
PR: feature-57-sell-resource

ignatIgnatov:


src/main/java/jewellery/inventory/dto/request/ProductDiscountRequestDto.java
VladoKat, line 1: Any reason to change this file?

src/main/java/jewellery/inventory/dto/request/PurchasedResourceInUserRequestDto.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+@SuperBuilder
+public class PurchasedResourceInUserRequestDto {
+
+  @NotNull private ResourceQuantityRequestDto resource;
```
VladoKat: Should we name this resourceAndQuantity?

src/main/java/jewellery/inventory/dto/response/PurchasedResourceInUserResponseDto.java
```diff
@@ -0,0 +1,20 @@
...
+
+@Getter
+@Setter
+@ToString
+@EqualsAndHashCode
+@SuperBuilder
+@NoArgsConstructor
+public class PurchasedResourceInUserResponseDto {
+
+  private ResourceQuantityResponseDto resource;
```
VladoKat: Should we rename to resourceAndQuantity

src/main/java/jewellery/inventory/dto/response/PurchasedResourceInUserResponseDto.java
```diff
@@ -0,0 +1,20 @@
...
+  private UserResponseDto owner;
```
VladoKat: I see we are using the PurchasedResourceInUserResponseDto when we return a list of all of the purchased resources. In this case we are going to have a lot of duplications of the user in the response. Does it make sense to create a wrapper around the List<PurchasedResourceInUserResponseDto> that we would name `PurchasedResourcesResponsesDto` and it would have a UserResponseDto and a `List<PurchasedResourceQuantityResponseDto>`.  `PurchasedResourceQuantityResponseDto` will have what `PurchasedResourceInUserResponseDto` currently has but without the `UserResponseDto`. I realize this got long and confusing. Ping me to discuss.

src/main/java/jewellery/inventory/dto/response/SaleResponseDto.java
```diff
@@ -19,6 +19,7 @@ public class SaleResponseDto {
   private UserResponseDto seller;
   private UserResponseDto buyer;
   private List<ProductResponseDto> products;
+  private List<PurchasedResourceInUserResponseDto> resources;
```
VladoKat: If we implement the suggested above, then this would be `List<PurchasedResourceQuantityResponseDto>` and it won't include the user multiple times.

src/main/java/jewellery/inventory/dto/response/SaleResponseDto.java
```diff
@@ -19,6 +19,7 @@ public class SaleResponseDto {
   private UserResponseDto seller;
   private UserResponseDto buyer;
   private List<ProductResponseDto> products;
```
VladoKat, line 22: I noticed we don't include the individual discounts we've provided for the Products in the SaleResponseDto. I realize this is not part of this ticket but we may want to add it.

src/main/java/jewellery/inventory/dto/response/resource/ResourceReturnResponseDto.java
```diff
@@ -0,0 +1,18 @@
...
+
+@Getter
+@Setter
+@SuperBuilder
+@RequiredArgsConstructor
+@AllArgsConstructor
+public class ResourceReturnResponseDto {
+    private ResourceResponseDto returnedResource;
+    private SaleResponseDto saleAfter;
+    private LocalDate date;
```
VladoKat: Are we adding the date just because of the date in the System events logging? I think we might not need this as we have the execution date in the events.

src/main/java/jewellery/inventory/dto/request/SaleRequestDto.java
```diff
@@ -20,8 +19,8 @@
 public class SaleRequestDto {
   @NotNull private UUID sellerId;
   @NotNull private UUID buyerId;
-
-  @Valid @NotEmpty private List<ProductDiscountRequestDto> products;
+  @Valid private List<PurchasedResourceInUserRequestDto> resources;
+  @Valid private List<ProductDiscountRequestDto> products;
```
VladoKat: Do we have a check somewhere now that at least one of those is not an empty list?

src/main/java/jewellery/inventory/model/User.java
```diff
@@ -54,6 +54,9 @@ public class User implements UserDetails {
   @OneToMany(mappedBy = "owner", cascade = CascadeType.ALL, orphanRemoval = true)
   private List<ResourceInUser> resourcesOwned = new ArrayList<>();
 
+  @OneToMany(mappedBy = "owner", cascade = CascadeType.ALL)
```
VladoKat: should we add orpahnRemoval here as well?

src/main/java/jewellery/inventory/model/resource/Resource.java
```diff
@@ -8,6 +8,7 @@
 import lombok.Data;
 import lombok.NoArgsConstructor;
 import lombok.experimental.SuperBuilder;
+import org.apache.commons.lang3.builder.ToStringExclude;
```
VladoKat: It's either obsolete, or we should mark the userAffiliation and productAffiliations with ToStringExclude but it should be coming from lombok instead of from apache.commons I think.

src/main/java/jewellery/inventory/service/SaleService.java
```diff
@@ -164,10 +180,117 @@ private void deleteSaleIfProductsIsEmpty(Sale sale) {
...
     }
     return productService.getProductReturnResponseDto(
         saleMapper.mapEntityToResponseDto(sale), productToReturn);
   }
+
+  private ResourceReturnResponseDto validateSaleAfterReturnResource(
+      Sale sale, PurchasedResourceInUser resourceToReturn) {
+    if (sale.getResources().isEmpty() && sale.getProducts().isEmpty()) {
+      return saleMapper.mapToResourceReturnResponseDto(resourceToReturn, null);
+    }
```
VladoKat: We can do the following. We'd need to add a null check inside saleMapper::mapEntityToResponseDto(Sale sale) to not throw NullPointer
```suggestion
    if (sale.getResources().isEmpty() && sale.getProducts().isEmpty()) {
      sale = null
    }
```

src/main/java/jewellery/inventory/service/SaleService.java
```diff
@@ -164,10 +180,117 @@ private void deleteSaleIfProductsIsEmpty(Sale sale) {
...
+                saleRequestDto.getResources().get(i).getResource().getResourceId());
+        purchasedResourceInUser.setResource(resource);
+        purchasedResourceInUser.setSalePrice(
+            resource
+                .getPricePerQuantity()
+                .multiply(saleRequestDto.getResources().get(i).getResource().getQuantity()));
+        purchasedResourceInUser.setDiscount(saleRequestDto.getResources().get(i).getDiscount());
+        purchasedResourceInUser.setQuantity(
+            saleRequestDto.getResources().get(i).getResource().getQuantity());
+        resources.add(purchasedResourceInUser);
```
VladoKat: Isn't this a mapper method? We'd need to pass the resource but otherwise it looks it could be part of the SalesMapper?

src/main/resources/db/migration/dev/V1_1__drop_existed_schema.sql
VladoKat, line 1: we shouldn't change the old versions. we won't be able to apply this without deleting the whole flyway table and therefore the whole DB.

src/main/resources/db/migration/prod/V1_3__create_product_price_discount_Table.sql
VladoKat, line 1: V1_3 and V1_4 seem to be in the opposite order in dev

src/test/java/jewellery/inventory/helper/SaleTestHelper.java
```diff
@@ -110,6 +181,63 @@ private static ProductResponseDto createProductResponseDto(UserResponseDto owner
...
+  }
+
+  private static ResourceResponseDto createResourceResponseDto() {
+    return ResourceResponseDto.builder()
+        .id(UUID.randomUUID())
+        .clazz("Pearl")
+        .quantityType("Gram")
+        .pricePerQuantity(getBigDecimal("20"))
+        .build();
+  }
```
VladoKat: pretty sure we have methods for these in ResourceTestHelper

src/test/java/jewellery/inventory/integration/AuthenticatedIntegrationTestBase.java
```diff
@@ -51,6 +52,7 @@ abstract class AuthenticatedIntegrationTestBase {
   void setup() {
     deleteAllImages();
     productRepository.deleteAll();
+    purchasedResourceInUserRepository.deleteAll();
```
VladoKat, line 55: If we go for the orphanRemoval property in the Sale entity, we should be able to delete all the purchasedResources through the sales.

src/test/java/jewellery/inventory/integration/SaleCrudIntegrationTest.java
VladoKat, line 1: Can we create a sale without both products and resources?

src/test/java/jewellery/inventory/unit/service/ResourceServiceTest.java
VladoKat, line 1: obsolete changes?

src/main/java/jewellery/inventory/service/SaleService.java
```diff
@@ -164,10 +180,117 @@ private void deleteSaleIfProductsIsEmpty(Sale sale) {
...
+
+  private void setFieldsOfResourcesAfterSale(Sale sale) {
+    List<PurchasedResourceInUser> resources = sale.getResources();
+    resources.forEach(
+        resource -> {
+          resource.setPartOfSale(sale);
+          resource.setOwner(sale.getBuyer());
+          resource.setSalePrice(
+              resource.getQuantity().multiply(resource.getResource().getPricePerQuantity()));
+        });
```
VladoKat, line 217: We are not setting discount here. Is it set somewhere else? I saw we are not setting it for the products as well.

VladoKat (changes requested):
Let me know what you think

src/main/java/jewellery/inventory/dto/request/ProductDiscountRequestDto.java
ignatIgnatov: Yes, when I sell only a resource, then product will be null

src/main/java/jewellery/inventory/dto/request/PurchasedResourceInUserRequestDto.java
ignatIgnatov: Ok, its Done!

src/main/java/jewellery/inventory/dto/response/PurchasedResourceInUserResponseDto.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/dto/request/ProductDiscountRequestDto.java
VladoKat: but the whole ProductDiscountRequest will be null, not just the productId

src/main/java/jewellery/inventory/dto/response/PurchasedResourceInUserResponseDto.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/dto/response/SaleResponseDto.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/SaleService.java
ignatIgnatov: I set discount in the mapper

src/main/java/jewellery/inventory/dto/response/resource/ResourceReturnResponseDto.java
ignatIgnatov: I removed the locale date from the response.

src/main/java/jewellery/inventory/dto/request/SaleRequestDto.java
ignatIgnatov: I think Yes. We can make a sale with only resources or only products, also both...

src/main/java/jewellery/inventory/dto/request/ProductDiscountRequestDto.java
ignatIgnatov: OK, I tested it and reverted to the old version

src/main/java/jewellery/inventory/model/User.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/model/resource/Resource.java
ignatIgnatov: This is some unusable import. I removed it.

src/main/resources/db/migration/prod/V1_3__create_product_price_discount_Table.sql
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/SaleCrudIntegrationTest.java
ignatIgnatov: I added an additional check with custom exception.

src/test/java/jewellery/inventory/helper/SaleTestHelper.java
ignatIgnatov: Yes that's right, we have a similar method. I refactored, but I think my code looked cleaner before...

src/main/resources/db/migration/dev/V1_1__drop_existed_schema.sql
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/SaleService.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/service/SaleService.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/integration/AuthenticatedIntegrationTestBase.java
ignatIgnatov: Done!

src/test/java/jewellery/inventory/helper/SaleTestHelper.java
ignatIgnatov: After second refactoring all is OK!

src/test/java/jewellery/inventory/unit/service/ResourceServiceTest.java
ignatIgnatov: Done!

src/main/java/jewellery/inventory/exception/sale/SaleImpossibleException.java
VladoKat, line 1: Perhaps EmptySaleException is a bit more specific?

src/main/java/jewellery/inventory/service/SaleService.java
```diff
@@ -40,43 +49,31 @@ public List<SaleResponseDto> getAllSales() {
   @LogCreateEvent(eventType = EventType.SALE_CREATE)
   @Transactional
   public SaleResponseDto createSale(SaleRequestDto saleRequestDto) {
+
+    throwExceptionWhenNoResourcesAndProductsInRequest(saleRequestDto);
```
VladoKat: Let's keep the throw method names consistent -> throwExceptionIfNoResourcesAndProductsInRequest

src/main/java/jewellery/inventory/service/SaleService.java
```diff
@@ -141,19 +163,23 @@ private void updateProductOwnersAndSale(
...
+          .toList();
+    }
+    return new ArrayList<>();
   }
 
-  private void deleteSaleIfProductsIsEmpty(Sale sale) {
-    if (sale.getProducts().isEmpty()) {
+  private void deleteSaleIfProductsAndResourcesAreEmpty(Sale sale) {
+    if (sale.getProducts().isEmpty() && sale.getResources().isEmpty()) {
       logger.info("Deleting sale with ID: {} since the products list is empty.", sale.getId());
```
VladoKat: We should also change the logger message to point out that both  resources and products are empty

src/test/java/jewellery/inventory/helper/ResourceTestHelper.java
```diff
@@ -69,6 +69,21 @@ public static Resource getPearl() {
...
+            .quantityType(QUANTITY_TYPE_UNIT)
+            .type(TYPE_PEARL)
+            .size(SIZE)
+            .quality(QUALITY)
+            .color(COLOR_PEARL)
+            .shape(SHAPE_PEARL)
+            .pricePerQuantity(price)
+            .note(NOTE)
+            .build();
+  }
```
VladoKat: Now there is code repetition. Can you edit the other method to be:
```
public static Resource getPearl() {
    return getPearl(PRICE_PER_QUANTITY);
}
```

ignatIgnatov: done!

src/main/java/jewellery/inventory/service/SaleService.java
ignatIgnatov: done!

src/main/java/jewellery/inventory/service/SaleService.java
ignatIgnatov: done!

src/main/java/jewellery/inventory/exception/sale/SaleImpossibleException.java
ignatIgnatov: done!

VladoKat (approved):

PR #87, opened 2024-02-22
BE product calculated price 58 - fix additionalPrice in Product

berki0:


src/main/java/jewellery/inventory/dto/request/ProductRequestDto.java
```diff
@@ -19,7 +19,7 @@ public class ProductRequestDto {
   @NotEmpty @Valid List<ResourceQuantityRequestDto> resourcesContent;
   private List<UUID> productsContent;
   private String description;
-  @NotNull private BigDecimal additionalPrice;
+  private BigDecimal additionalPrice;
```
VladoKat: Why remove the NotNull validation?

src/test/java/jewellery/inventory/helper/ProductTestHelper.java
```diff
@@ -50,7 +50,7 @@ public static ProductRequestDto getProductRequestDto(
     productRequestDto.setCatalogNumber("1111");
     productRequestDto.setProductionNumber("1234");
     productRequestDto.setDescription("Test");
-    productRequestDto.setAdditionalPrice(BigDecimal.TEN);
+    productRequestDto.setAdditionalPrice(null);
```
VladoKat: I guess having an additional price breaks the tests. You should either take it into account when asserting for the end price or make it 0. Let's not allow null as an option

VladoKat (changes requested):

VladoKat (approved):

VladoKat:
Related ticket https://trello.com/c/QGlzAOpu/58-be-product-calculated-price

PR #90, opened 2024-02-25
feat-65: change organization permissions persistence

VladoKat:
This makes it so we save the permissions in the DB with their enum values.
https://trello.com/c/O9cEbN26/65-be-expose-organization-create-read-endpoint


Alekseeybg (approved):
Very sophisticated fix. I like it. Approved.

PR #97, opened 2024-03-09
Fixed update owner bug

LuboKar:


src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
```diff
@@ -328,7 +327,7 @@ void testUpdateProductSuccessfullyWhenProductIsPartOfAnotherProduct() {
     productService.updateProduct(innerProduct.getId(), productRequestDto);
 
     verify(productRepository, times(1)).findById(innerProduct.getId());
-    verify(userRepository,times(2)).findById(user.getId());
```
VladoKat, line 331: I guess this was not supposed to be 2 in the first place?

src/test/java/jewellery/inventory/integration/ProductCrudIntegrationTest.java
```diff
@@ -389,6 +389,32 @@ void updateProductSuccessfully() throws JsonProcessingException {
...
+
+    ProductResponseDto productResponseDto = response.getBody();
+
+    Map<String, Object> expectedEventPayload =
+        getUpdateEventPayload(product, productResponseDto, objectMapper);
+
+    systemEventTestHelper.assertEventWasLogged(PRODUCT_UPDATE, expectedEventPayload);
+
+    assertEquals(oldOwnerId, productResponseDto.getOwner().getId());
+    assertNotEquals(productRequestDto.getOwnerId(), productResponseDto.getOwner().getId());
```
VladoKat: Restructure to 3 groups separated by an empty line: setup, action, assertions.

VladoKat (changes requested):

src/test/java/jewellery/inventory/unit/service/ProductServiceTest.java
LuboKar: We were looking for a user in the database because we needed the user to be passed to setProductField and since we changed it to product.owner() we no longer call it 2 times.

VladoKat (approved):

PR #98, opened 2024-03-11
Be expose organization destruction endpoint 70

berki0:
https://trello.com/c/uSzghBbV/70-be-expose-organization-destruction-endpoint


src/main/java/jewellery/inventory/controller/OrganizationController.java
```diff
@@ -64,6 +64,13 @@ public void deleteUserInOrganization(
     userInOrganizationService.deleteUserInOrganization(userId, organizationId);
   }
 
+  @Operation(summary = "Delete a organization")
```
VladoKat:
```suggestion
  @Operation(summary = "Delete an organization")
```

src/main/java/jewellery/inventory/service/OrganizationService.java
```diff
@@ -67,6 +80,37 @@ private List<Organization> getAll() {
...
+        currentUser.getId(),
+        organization.getId());
+  }
+
+  private boolean hasPermission(
+      User user, Organization organization, OrganizationPermission permission) {
+    return organization.getUsersInOrganization().stream()
+        .anyMatch(
+            userInOrganization ->
+                userInOrganization.getUser().equals(user)
```
CodeLikeAlexito: Here we can have NPE. Better to write it as:
user.equals(userInOrganization.getUser())

src/main/java/jewellery/inventory/service/OrganizationService.java
```diff
@@ -67,6 +80,37 @@ private List<Organization> getAll() {
...
+  private void validateOrganizationProductsAndResources(Organization organization) {
+    if (!organization.getProductsOwned().isEmpty()) {
+      throw new OrganizationProductsException(organization.getId());
+    }
+    if (!organization.getResourceInOrganization().isEmpty()) {
+      throw new OrganizationResourcesException(organization.getId());
+    }
+  }
+
+  public void validateCurrentUserPermission(
```
CodeLikeAlexito, line 108: Why this is public? Do we need it outside of this class ?

CodeLikeAlexito, line 108: When having validation methods, it is good practice to return type (boolean for example). Probably better to return boolean and if it doesn't throw exception to return true.
Or the other option is to return false if it is not valid and to throw the exception from where validation is called.

src/main/java/jewellery/inventory/service/OrganizationService.java
```diff
@@ -43,6 +47,15 @@ public OrganizationResponseDto create(OrganizationRequestDto organizationRequest
     return organizationMapper.toResponse(organization);
   }
 
+  @LogDeleteEvent(eventType = EventType.ORGANIZATION_DELETE)
+  public void delete(UUID organizationId) {
```
CodeLikeAlexito, line 56: Here also, probably it is better to return something and not just void method. If something happens and for some reason we can't delete we can return for example boolean or our custom object. And then when you call this from controller, based on result you can return either: For positive deletion ( NO CONTENT) and for negative ( NOT FOUND or something else)
This is just a tip to consider

src/main/java/jewellery/inventory/service/OrganizationService.java
berki0: yes we use them in UserInOrganizationService

src/main/java/jewellery/inventory/exception/organization/OrganizationProductsException.java
```diff
@@ -0,0 +1,9 @@
+package jewellery.inventory.exception.organization;
+
+import java.util.UUID;
+
+public class OrganizationProductsException  extends  RuntimeException{
```
VladoKat: Let's rename the exception `OrphanProductsInOrganizationException`

src/main/java/jewellery/inventory/exception/organization/OrganizationResourcesException.java
```diff
@@ -0,0 +1,8 @@
+package jewellery.inventory.exception.organization;
+import java.util.UUID;
+
+public class OrganizationResourcesException extends RuntimeException {
```
VladoKat: Same here: OrphanResourcesInOrganizationException

src/main/java/jewellery/inventory/service/OrganizationService.java
VladoKat: So far we've made delete requests return nothing if the request is valid and the actions is successful. If there is a problem with the deletion, we are creating the response in the GlobalExceptionHandler as a result of throwing an exception.

src/main/java/jewellery/inventory/service/OrganizationService.java
```diff
@@ -49,6 +52,15 @@ public OrganizationResponseDto create(OrganizationRequestDto organizationRequest
...
+    Organization organizationForDelete = getOrganization(organizationId);
+    validateCurrentUserPermission(
+        organizationForDelete, OrganizationPermission.DESTROY_ORGANIZATION);
+    validateOrganizationProductsAndResources(organizationForDelete);
```
VladoKat:
```suggestion
    verifyNoProductsOrResourcesInOrganization(organizationForDelete);
```

src/main/java/jewellery/inventory/service/OrganizationService.java
VladoKat: I don't quite agree. We are using the validate methods to throw exception if the business logic doesn't allow us to continue with the given operation. This sounds reasonable to me as we are interrupting the current logic from continuing and are directly creating a response out of this failed validation. What you're suggesting will do the same but will add more complexity to the code due to additional if statements. The only benefit I see is if we want to include multiple error messages to the response. In this ticket it could make sense if the organization has both products and resources available when trying to delete it. This is an actual case here but I don't think it's worth it to explore it deeper for now.

src/test/java/jewellery/inventory/unit/service/UserInOrganizationServiceTest.java
VladoKat, line 1: Obsolete change here.

src/test/java/jewellery/inventory/unit/service/OrganizationServiceTest.java
VladoKat, line 1: We should have a test for trying to delete an organization and the user doesn't have permissions to do so (no DESTROY_ORGANIZATION permission)

VladoKat (changes requested):

src/test/java/jewellery/inventory/helper/OrganizationTestHelper.java
```diff
@@ -24,13 +24,21 @@ public static Organization getTestOrganization() {
     return organization;
   }
 
+  public static Organization addUserWithNoPermissions(Organization organization, User user) {
```
VladoKat: Let's keep the naming consistent with the other similar method
```suggestion
  public static Organization getOrganizationWithUserWithNoPermissions(Organization organization, User user) {
```

VladoKat (approved):

PR #105, opened 2024-04-07
adding failed test cases to ProductInOrganizationServiceTest

berki0:


LuboKar (commented):
I'm not sure if the assertions in "successful" tests are necessary because you are stubbing the mapper , which means all the responses will be the same every time, so the assertions will always be true.

src/test/java/jewellery/inventory/unit/service/ProductInOrganizationServiceTest.java
VladoKat, line 1: There are some unused mocked fields (productRepository, resourceInProductRepository)

VladoKat: Also, there seems to be no empty lines between the tests on the bottom. Please, format the file

VladoKat (changes requested):

VladoKat (approved):

PR #110, opened 2024-04-16
add product authors validation

berki0:
adding validation method for products authors if they are in organization


src/main/java/jewellery/inventory/service/ProductInOrganizationService.java
```diff
@@ -168,6 +171,21 @@ private void setProductFields(
     logger.debug("Product fields have been set successfully for product: {}", product);
   }
 
+  private void validateProductAuthors(Organization organization, List<User> authors) {
```
VladoKat: This method checks if the authors are part of the owning organization. Let's name it `validateUsersAreMembersOfOrganization`

VladoKat (changes requested):

VladoKat (approved):

PR #135, opened 2025-12-04
Feature add SKU and update resource entities

LuboKar:


src/main/java/jewellery/inventory/dto/request/resource/AllowedValueRequestDto.java
```diff
@@ -12,4 +12,5 @@ public class AllowedValueRequestDto {
   @NotEmpty private String resourceClazz;
   @NotEmpty private String fieldName;
   @NotEmpty private String value;
+  private String sku;
```
VladoKat, line 15: Should we allow empty sku in the request?

src/main/java/jewellery/inventory/dto/request/resource/ColoredStoneMeleeRequestDto.java
```diff
@@ -0,0 +1,23 @@
...
+public class ColoredStoneMeleeRequestDto extends ResourceRequestDto {
+  private String color;
+  private String cut;
+  private String clarity;
+  private String shape;
+  private String size;
+  private String type;
+  private BigDecimal carat;
+  private String colorHue;
+  private String treatment;
```
VladoKat: Should we force some validation based on the required fields?

src/main/java/jewellery/inventory/dto/response/resource/ColoredStoneResponseDto.java
```diff
@@ -0,0 +1,24 @@
...
+@ToString(callSuper = true)
+@EqualsAndHashCode(callSuper = true)
+@SuperBuilder
+@NoArgsConstructor
+public class ColoredStoneResponseDto extends ResourceResponseDto {
+  private String color;
+  private BigDecimal carat;
+  private String cut;
+  private String clarity;
+  private String size;
```
VladoKat: I remember we discussed but don't remember the decision about whether we should return the size as dimensionX, Y, Z and delegate the formatting to XxYxZ to the frontend.

src/main/resources/db/migration/dev/V1_11_1__update_resource_field_values.sql
VladoKat, line 1: Most of this migration needs to be applied to both dev and prod. Please, move the general tables structure updates to common and create another migration to update the records that are specific to dev.

src/test/java/jewellery/inventory/integration/ResourceInOrganizationCrudIntegrationTest.java
```diff
@@ -362,8 +364,17 @@ private OrganizationResponseDto createOrganization() {
     return organizationResponseDto;
   }
 
-  private ResourceResponseDto createResourceResponse() {
-    ResourceRequestDto resourceRequest = getPearlRequestDto();
+  private ResourceResponseDto createResourceResponse(String resourceClazz) {
+    ResourceRequestDto resourceRequest =
+        switch (resourceClazz) {
+          case "Pearl" -> getPearlRequestDto();
+          case "Metal" -> getMetalRequestDto();
```
VladoKat: Can we use the constants you created above for those? Also, are we ever using the other clazz types? Do we need to have them if we don't? If there are tests where it makes sense we want to try with all of the resources, we should implement some ParameterizedTests and execute them with all resource types. Use your best judgement about that. We can discuss if it is unclear.

VladoKat (changes requested):

VladoKat (approved):

PR #136, opened 2026-01-07
Feature remove obsolete endpoints

LuboKar:


src/main/java/jewellery/inventory/controller/OrganizationController.java
```diff
@@ -32,6 +33,7 @@ public List<OrganizationResponseDto> getAllOrganizationsForCurrentUser() {
     return organizationService.getAllOrganizationsResponsesForCurrentUser();
   }
 
+  @NotUsedYet(reason = "Pending frontend implementation")
```
VladoKat, line 37: How is the annotation different than just adding a comment that it is not used?

src/main/java/jewellery/inventory/controller/ProductController.java
VladoKat, line 1: Now that I see the deleted methods, I think it is more meaningful if we put all of the methods on endpoints 
_/organizations/products/*_ to be just _/products/*_ and therefore be moved in this controller. Let's leave it as is right now but I would like to discuss it to see how the structure makes more sense.

src/main/java/jewellery/inventory/controller/SaleController.java
VladoKat, line 1: Same could be said about the Sales. A sale is defined by the seller (organization) and buyer (user) so it kind of makes sense to be something on its own and not have it as part of the organizations. I believe we made the decision to put it under organizations because we had duplicate functionalities. But now we could use just /sales as endpoint. Let's discuss it together with the product topic.

src/main/java/jewellery/inventory/service/OrganizationSaleService.java
```diff
@@ -29,45 +38,46 @@ public class OrganizationSaleService {
...
   private static final Logger logger = LogManager.getLogger(OrganizationSaleService.class);
 
-  private final SaleService saleService;
   private final SaleMapper saleMapper;
   private final OrganizationService organizationService;
   private final UserService userService;
   private final ResourceInOrganizationService resourceInOrganizationService;
   private final SaleRepository saleRepository;
   private final ProductService productService;
+  private final PurchasedResourceInUserRepository purchasedResourceInUserRepository;
```
VladoKat: We should use the service instead of the repository for the purchasedResource. This will keep the consistency that a service should only use a repo for the model they are representing and any actions related to other models should happen via service calls.

VladoKat (changes requested):
Nothing major to be done. Let's discuss before merging though.

VladoKat (approved):

VladoKat (approved):

VladoKat (approved):

PR #138, opened 2026-01-28
feature: add related system events table

LuboKar:


src/main/java/jewellery/inventory/dto/response/SystemEventLiteResponseDto.java
```diff
@@ -0,0 +1,26 @@
...
+  private UUID id;
+  private Instant timestamp;
+  private EventType type;
+  private Map<String, Object> executor;
+
+  public SystemEventLiteResponseDto(UUID id, Instant timestamp, EventType type, Object executor) {
+    this.id = id;
+    this.timestamp = timestamp;
+    this.type = type;
+    this.executor = (Map<String, Object>) executor;
```
VladoKat, line 24: I imagine RequiredArgsConstructor doesn't work because of this?

src/main/java/jewellery/inventory/service/SystemEventService.java
```diff
@@ -65,4 +82,52 @@ private Map<String, Object> getCurrentUser() {
...
+
+    if (currentNode.isObject()) {
+      var fieldIterator = currentNode.fields();
+
+      while (fieldIterator.hasNext()) {
+        var field = fieldIterator.next();
+        String fieldName = field.getKey();
+        JsonNode fieldValue = field.getValue();
+
+        if (("id".equals(fieldName) || "contentOf".equals(fieldName))
```
VladoKat: do we save UUIDs only under "id" and "contentOf"? That looks very flaky. I see in the responses we have productId, partOfSale that won't be caught by this. I think we shouldn't rely on field names but rather try to find all values that are UUIDs.

src/test/java/jewellery/inventory/helper/SystemEventTestHelper.java
```diff
@@ -32,15 +31,28 @@ public class SystemEventTestHelper {
...
-    Optional<SystemEvent> event =
-        findEventByTypeAndEntity(testRestTemplate, objectMapper, eventType, toCompare);
+    Optional<SystemEventLiteResponseDto> eventOptional =
+        findEventByTypeAndEntity(testRestTemplate, objectMapper, eventType, toCompare, relatedId);
+
+    SystemEventLiteResponseDto event = null;
+
+    if (eventOptional.isPresent()) {
+      event = eventOptional.get();
+    }
```
VladoKat: If event is not present, we shouldn't continue if I understand correctly. I would replace the if with 
```suggestion
    assertTrue(eventOptional.isPresent()) 
    event = eventOptional.get();
```
Then the assertTrue for the same below is obsolete

src/test/java/jewellery/inventory/helper/SystemEventTestHelper.java
```diff
@@ -58,26 +70,31 @@ public static Map<String, Object> getCreateOrDeleteEventPayload(
...
 
-    ResponseEntity<String> response = testRestTemplate.getForEntity("/system-events", String.class);
-    return objectMapper.readValue(response.getBody(), new TypeReference<>() {});
+    ResponseEntity<List<SystemEventLiteResponseDto>> response =
+        testRestTemplate.exchange(
+            "/system-events/related-to/" + relatedId,
+            HttpMethod.GET,
+            null,
+            new ParameterizedTypeReference<List<SystemEventLiteResponseDto>>() {});
+    return response.getBody();
```
VladoKat: This method is named fetchAllSystemEvents but is fetching the relatedTo events. Does it have the same meaning still? The objectMapper and the throws statement seem to be obsolete now.
Also the method that uses it (findEventByTypeAndEntity) has an obsolete parameter "toCompare".

src/test/java/jewellery/inventory/integration/SystemEventCrudIntegrationTest.java
```diff
@@ -1,53 +1,137 @@
...
+    List<SystemEventLiteResponseDto> events = eventResponse.getBody();
     assertFalse(events != null && events.isEmpty());
     assertNotNull(events, "Retrieved events list is empty");
+    assertEquals(EventType.USER_CREATE, events.getFirst().getType());
+  }
 
-    Map<String, Object> expectedEventPayload =
-        getCreateOrDeleteEventPayload(detailedUserResponseDto, objectMapper);
+  @Test
+  void testGetSystemEventShouldThrow() {
```
VladoKat, line 50: I wouldn't say this is throwing. It is returning that the id is not found. Maybe rename to:
```suggestion
  void testGetSystemEventNotFound() {
```

src/test/java/jewellery/inventory/integration/SystemEventCrudIntegrationTest.java
```diff
@@ -1,53 +1,137 @@
...
+
   private DetailedUserResponseDto createAndSaveUser() {
     UserRequestDto userRequest = createTestUserRequest();
 
     return this.testRestTemplate
         .postForEntity("/users", userRequest, DetailedUserResponseDto.class)
         .getBody();
   }
+
+  private ResponseEntity<List<SystemEventLiteResponseDto>> getRelatedEventsQuery(UUID id) {
```
VladoKat: I don't think this is a query. I would suggest:

```suggestion
  private ResponseEntity<List<SystemEventLiteResponseDto>> getRelatedEventsResponse(UUID id) {
```

src/test/java/jewellery/inventory/unit/service/SystemEventServiceTest.java
VladoKat, line 1: Rename to "SystemEventServiceTest" as this is what is under test.

src/test/java/jewellery/inventory/unit/service/SystemEventServiceTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    assertNotNull(event);
+    assertEquals(event, systemEventResponseDto);
+
+    verify(systemEventRepository, times(1))
+        .findByIdWithoutRelatedIds(systemEventResponseDto.getId());
+  }
+
+  @Test
+  void testGetEventsRelatedToShouldReturnEmptyList() {
```
VladoKat, line 89: What is the difference for not found exception for a UUID and an empty list for a UUID? Do we somehow make sure that the UUID we are searching for is used somewhere in our system?

VladoKat (changes requested):
There are a few small things that need to be addressed but the main thing is the way we are gathering the relatedIds in the system event service. It should be more general so we don't miss some ids.

VladoKat (approved):

PR #144, opened 2026-06-11
feature: add system permissions

LuboKar:


VladoKat (changes requested):
AI Review:
Findings
1. Allowed-value endpoints still bypass the new system RBAC.
AllowedValueController.java (line 18) exposes global POST, DELETE, and GET operations without @PreAuthorize, and AllowedValueService.java (line 18) does not enforce authorization either. These endpoints are not organization-scoped and they mutate/read shared resource metadata, so users with the old global ADMIN app role but no new system permissions can still manage them because SecurityConfig.java (line 79) only requires ROLE_ADMIN globally. I would gate these with system permissions, likely SYSTEM_RESOURCE_READ for GET and either existing resource write permissions or dedicated allowed-value permissions for create/delete. Please also add denied integration tests for users without those system permissions.

2. Invalid mixed-scope role definitions can return 500 instead of a client error.
ScopedRole.java (line 37) validates role type vs permission scope in a JPA callback and throws IllegalStateException. That exception is not mapped by GlobalExceptionHandler.java (line 65), so creating a SYSTEM role with organization permissions, or the reverse, can surface as an internal server error. This should be validated in the service/API path and converted to a 400-level response with test coverage.

3. Role-membership scope invariants are not protected at the database/native insert layer.
RoleMembership.java (line 34) enforces “system role has no organization” and “organization role has organization” only through entity callbacks, but RoleMembershipRepository.java (line 104) inserts memberships with a native query, bypassing those callbacks. The unique constraint at RoleMembership.java (line 12) also includes nullable organization_id, so duplicate system memberships for the same user/role are not reliably prevented in Postgres. I’d add DB-level constraints/partial unique indexes or split the insert paths so system and organization membership writes cannot create invalid state.

Overall, the direction looks coherent: user/resource/organization/role endpoints are mostly wired into system or organization permissions, and system-event filtering is handled in the service where mixed visibility makes sense. The main gap is consistency: every non-organization global endpoint should either have a system permission or be explicitly documented as intentionally self-readable.


My review:
I agree with point 2 and 3. Not sure I agree with 1 since we don't care about allowed values protection I guess.

VladoKat (approved):

